PR #1, opened 2023-07-23
Fix streaming

sumo43:
I'm not sure if there is a way to get sentencepiece to respect spacing when decoding individual tokens, but here is a way to have streaming working correctly (at the cost of being slightly slower).


karpathy:
Okay I see. Haha.

karpathy:
Okay I'll take it! Thank you.

PR #8, opened 2023-07-23
Docs: Compile with `-O3`

ashvardanian:
Amazing project, but it's even better than reported, if you enable `-O3`. On M2 Pro I got from 18 toks/s:

```txt
achieved tok/s: 18.328490461200566
```

to 

```txt
achieved tok/s: 96.74965746530997
```

Thanks for sharing!


karpathy:
Wow that's awesome ty, was def on my todo to try, didn't expect as much! :)

karpathy:
I made a more accurate diff to readme and committed, ty for the heads up!

Issue #9, opened 2023-07-23
SentencePieceProcessor vs C [Enhancement]


krzysztof-jusiak:
To do the inference with just c and without sentence piece processor one easy way would be to save the id to token in the model.bin?

```
tokenizer = SentencePieceProcessor(tokenizer_model)
vocab = [tokenizer.id_to_piece(id) for id in range(tokenizer.get_piece_size())]
```

and then just an array to get the proper token from id

```
auto decode(auto id) {
  return vocab[id];
}
```

That would allow not to use the run_wrap.py and it would be in pure C (kinda).

trholding:
Ref: https://github.com/google/sentencepiece/blob/master/src/spm_decode_main.cc

karpathy:
Closing old issue. This has been implemented in run.c for a while with ASCII, and UTF-8 support is about to merge soon in https://github.com/karpathy/llama2.c/pull/226 .

PR #12, opened 2023-07-23
Emscripten build (demo, quick and dirty)

ggerganov:
This is a very ugly hack to demonstrate build with Emscripten

Not intended to merge as it will just ruin the beauty of the implementation and most likely there is a much better way to do this. Mainly for educational purposes.

```bash
emcc -O3 run.c \
  -o web/llama2.js \
  -s EXPORTED_FUNCTIONS='["_main", "_main_loop", "_malloc", "_free"]' \
  -s EXPORTED_RUNTIME_METHODS='["ccall"]' \
  -s ALLOW_MEMORY_GROWTH=1 \
  --preload-file model.bin \
  --preload-file vocab.bin
```

The `vocab.bin` is generated in a similar way explained in #9 

The produced artifacts will be generated in the `web` subfolder.

Example: https://ggerganov.com/llama2.c

### DIFF
--- a/run.c
+++ b/run.c
@@ -383,80 +383,100 @@ int argmax(float* v, int n) {
 
 // ----------------------------------------------------------------------------
 
-int main(int argc, char *argv[]) {
+// poor man's C argparse
+char *checkpoint = NULL;
+float temperature = 0.9f;
+
+// read in the config header
+Config config;
+
+TransformerWeights weights;
+
+// create and init the application RunState
+RunState state;
+
+// the current position we are in
+int next;
+int token = 1; // 1 = BOS token in Llama-2 sentencepiece
+int pos = 0;
+
+char * vocab[32000];
+
+void main_loop(void * dummy) {
+    // forward the transformer to get logits for the next token
+    transformer(token, pos, &config, &state, &weights);
+
+    // sample the next token
+    if(temperature == 0.0f) {
+        // greedy argmax sampling
+        next = argmax(state.logits, config.vocab_size);
+    } else {
+        // apply the temperature to the logits
+        for (int q=0; q<config.vocab_size; q++) { state.logits[q] /= temperature; }
+        // apply softmax to the logits to get the probabilities for next token
+        softmax(state.logits, config.vocab_size);
+        // we now want to sample from this distribution to get the next token
+        next = sample(state.logits, config.vocab_size);
+    }
+    //printf("%d\n", next);
+    printf("%s\n", vocab[next]);
+    fflush(stdout);
+
+    // advance forward
+    token = next;
+    pos++;
+}
+
+#ifdef __EMSCRIPTEN__
+#include "emscripten.h"
+#endif
+
+int main() {
     setbuf(stdout, NULL); // disable stdout buffering
 
-    // poor man's C argparse
-    char *checkpoint = NULL;
-    float temperature = 0.9f;
-    // 'checkpoint' is necessary arg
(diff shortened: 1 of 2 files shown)


run.c
```diff
@@ -383,80 +383,100 @@ int argmax(float* v, int n) {
...
-            // apply softmax to the logits to get the probabilities for next token
-            softmax(state.logits, config.vocab_size);
-            // we now want to sample from this distribution to get the next token
-            next = sample(state.logits, config.vocab_size);
+    // model
+    {
+        checkpoint = "model.bin";
+        FILE *file = fopen(checkpoint, "rb");
+        if (!file) {
+            printf("Unable to open file!");
```
alexeykudinkin, line 447:
```suggestion
            printf("Unable to open file!");
            return 1;
```

run.c
```diff
@@ -383,80 +383,100 @@ int argmax(float* v, int n) {
...
     }
 
-    free_run_state(&state, &config);
-    free_weights(&weights, &config);
-    return 0;
+    // vocab
+    {
+        FILE *file = fopen("vocab.bin", "r");
+        if (!file) {
+            printf("Unable to open file!");
```
alexeykudinkin, line 461:
```suggestion
            printf("Unable to open file!");
            return 1;
```

alexeykudinkin (commented):
Worth merging IMO even if not for Emscripten, then for cleaning up the main sequence at the very least

karpathy:
Very cool ty! I'm not as familiar with emscripten, I'll take some time to look at this and whether there is a way to support it gracefully.

karpathy:
@ggerganov one thing I really like is the tokenizer inside C here, removing the need to use the run_wrap.py and sentencepiece. But it's not clear how you produced vocab.bin, presumably it's an export script?

karpathy:
> @ggerganov one thing I really like is the tokenizer inside C here, removing the need to use the run_wrap.py and sentencepiece. But it's not clear how you produced vocab.bin, presumably it's an export script?

nvm got it working in https://github.com/karpathy/llama2.c/commit/3bfa5665d1bd278041cf272bd46db19129d6c3fd

ggerganov:
Yes, for this PR I had exported the vocab from `llama.cpp`, which has been generated with a similar script as yours.

I'm not sure about the cause of the leading space - we have it in `llama.cpp` output as well and haven't dug to understand how to fix it yet.

kroggen:
The tokenizer stores a leading space on some tokens.

In python we use the `Decode` function, that removes the leading space on the first token

But this repo is using the `id_to_piece()` function to export the tokenizer

```python
python3
>>> import sentencepiece as spm
>>> sp = spm.SentencePieceProcessor(model_file='tokenizer.model')
>>> sp.id_to_piece([9038])
['▁Once']
```

Then it is simply printed, without considering if it is the first token:

https://github.com/karpathy/llama2.c/blob/f5650891d53e15605bb066a978c0182b12e91cd8/run.c#L471

gohai:
I recreated @ggerganov's changes against the current tree (with support for prompt): https://github.com/gohai/llama2.c/commits/emscripten

rahuldshetty:
I've built a JS framework around this idea to run Language Models on Web. 
This also leverages Emscripten to compile llama2.c into WASM JS.

llama2.c on Web Demo: https://rahuldshetty.github.io/ggml.js-examples/llama2_tinystories.html
ggml.js Framework: https://github.com/rahuldshetty/ggml.js

PR #10, opened 2023-07-23
Bigger number better with -funsafe-math-optimizations flag.

luigifcruz:
Enabling spicy math optimization `-funsafe-math-optimizations` yields ~6x performance uplift. 

Screenshot 2023-07-23 at 5 05 50 PM

PR #13, opened 2023-07-23
matmul: optionally use vDSP from macOS Accelerate Framework

jmccarthy:
Illustrating >500 tokens/s (M1 Max) by directly replacing matmul with vDSP.

All the `#ifdef` makes run.c a bit too ugly to merge, but fun to see it work!

### DIFF
--- a/run.c
+++ b/run.c
@@ -10,6 +10,18 @@ Then run with:
 $ ./run
 */
 
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#if TARGET_OS_MAC
+#include <Accelerate/Accelerate.h>
+#define ACCELERATE_AVAILABLE 1
+#else
+#define ACCELERATE_AVAILABLE 0
+#endif
+#else
+#define ACCELERATE_AVAILABLE 0
+#endif
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <time.h>
@@ -221,15 +233,22 @@ void softmax(float* x, int size) {
     }
 }
 
-void matmul(float* xout, float* x, float* w, int n, int d) {
+void matmul(float *xout, float *x, float *w, int n, int d)
+{
     // W (d,n) @ x (n,) -> xout (d,)
-    for (int i = 0; i < d; i++) {
+#if ACCELERATE_AVAILABLE
+    vDSP_mmul(w, 1, x, 1, xout, 1, d, 1, n);
+#else
+    for (int i = 0; i < d; i++)
+    {
         float val = 0.0f;
-        for (int j = 0; j < n; j++) {
+        for (int j = 0; j < n; j++)
+        {
             val += w[i * n + j] * x[j];
         }
         xout[i] = val;
     }
+#endif
 }
 
 void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights* w) {
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 ## llama2.c
 
 Have you ever wanted to inference a baby [Llama 2](https://ai.meta.com/llama/) model in pure C? No? Well, now you can!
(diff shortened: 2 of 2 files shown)


joey00072:
I didn’t see any speed improvement after ` -funsafe-math-optimizations `

source: https://github.com/karpathy/llama2.c/pull/10

ystolzenburg:
Even with `-funsafe-math-optimizations`, it's still ~150 tok/s faster with vDSP on my M1 Pro.
screenshot

bsharper:
On my Intel iMac, this change increases the tok/s from 20 to around 440 tok/s.

karpathy:
On my M1 Macbook this change doesn't create improvement for some reason.

bsharper:
Here are the performance differences I'm seeing (average of 5 runs):

**Unmodified `run`:**
- 15m average tok/s: 70.2 (std dev: 0.44)
- 44m average tok/s: 24.39 (std dev: 0.15)
- 110m average tok/s: 9.36 (std dev: 0.06)

**Modified `run` with changes in this PR:**
- 15m average tok/s: 280.48 (std dev: 6.51)
- 44m average tok/s: 109.91 (std dev: 2.37)
- 110m average tok/s: 40.96 (std dev: 0.73)

Computer is 2020 iMac w/ 3.3GHz 6‑core Intel i5, running on macOS 13.4.1c.

jmccarthy:
@karpathy or anyone else: if you think the Accelerate framework optimizations are worth including in some way, I can update the PR with a more comprehensive approach. 

IMHO, "use platform-specific matmul like vDSP" doesn't aid understanding or intuition, whereas the `#pragma omp` change achieves both performance _and_ understanding.

Will keep this open a bit longer in case anyone wants to weigh in, otherwise will close.

run.c
```diff
@@ -221,15 +233,22 @@ void softmax(float* x, int size) {
...
-    for (int i = 0; i < d; i++) {
+#if ACCELERATE_AVAILABLE
+    vDSP_mmul(w, 1, x, 1, xout, 1, d, 1, n);
+#else
+    for (int i = 0; i < d; i++)
+    {
         float val = 0.0f;
-        for (int j = 0; j < n; j++) {
+        for (int j = 0; j < n; j++)
+        {
```
kroggen, line 246: Why did you do this?
It would be way less changes if you had kept the original position of curly braces (on 3 places)

karpathy:
Are people with Apple Silicon macs able to reproduce a speed with this? My performance on my M1 MacBook Air actually goes down by almost a half. I add:

```
#include <Accelerate/Accelerate.h>
```

and then just use:

```
    vDSP_mmul(w, 1, x, 1, xout, 1, d, 1, n);
}
```

and compile with

```
clang -Ofast -framework Accelerate -o run run.c -lm
```

performance tanks. Is this just my setup somehow?

kroggen:
On my MacBook Intel there is no perceived difference in performance

karpathy:
I definitely *would* expect performance improvements out of using Apple Silicon GPU... probably this isn't done correctly just yet in this PR, have to dig into it.

bsharper:
Can confirm an average decrease in performance of around 20% on my M1 MBP when using this PR. Numbers below are averages of 5 runs.

**OS**: macOS 13.4.1 22F82 arm64
**CPU**: Apple M1
**GPU**: Apple M1

**Modified `run` with changes in this PR:**
- 15m average tok/s: 579.55 (std dev: 3.09)
- 44m average tok/s: 176.39 (std dev: 8.65)
- 110m average tok/s: 70.24 (std dev: 0.45)

**Unmodified `run`:**
- 15m average tok/s: 659.73 (std dev: 10.68)
- 44m average tok/s: 274.37 (std dev: 1.59)
- 110m average tok/s: 104.62 (std dev: 1.31)

karpathy:
This PR doesn't work :( closing.

PR #15, opened 2023-07-23
Tokenizer in C

python273:
Very simple. It doesn't remove leading spaces. Might be good idea to test with weird unicode?

Screenshot 2023-07-24 at 01 06 16


karpathy:
Very interesting!! Will take a look tonight thank you!

python273:
Looks like #12 also has tokenizer implementation 😅

karpathy:
@python273 do you understand where #12 gets the vocab.bin file from? presumably a similar export script?

export_tokenizer.py
```diff
@@ -0,0 +1,43 @@
...
+token_blob = []
+offsets = []
+offset = 0
+for i in range(vocab_size):
+    t = sp_model.id_to_piece(i)
+    if i == bos_id:
+        t = '\n<s>\n'
+    elif i == eos_id:
+        t = '\n</s>\n'
+    elif len(t) == 6 and t.startswith('<0x') and t.endswith('>'):
```
karpathy, line 22: ??? some comments around here could be nice. i haven't dug into sp too much but this looks odd

karpathy:
@python273 I got it working. What are the leading spaces? How does SentencePiece handles this 🤔

karpathy:
Solved here I think.
https://github.com/karpathy/llama2.c/commit/3bfa5665d1bd278041cf272bd46db19129d6c3fd

ty!!

karpathy:
(I didn't solve the leading space issue... leaving it there for now, will look later into what sentencepeice does with this)

PR #17, opened 2023-07-23
add simple makefile

tmc:
This adds a simple makefile that gets going as straightforward as `make run-inference`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,30 @@
+# Contains helpful targets for building and running the project.
+
+all: build
+
+.PHONY: build
+build: run
+
+run: run.c
+	gcc -O3 -o run run.c -lm
+
+.PHONY: run-inference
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,30 @@
...
+all: build
+
+.PHONY: build
+build: run
+
+run: run.c
+	gcc -O3 -o run run.c -lm
+
+.PHONY: run-inference
+run-inference: .venv run out/model.bin
```
svmhdvn, line 12: I'd also probably add a `run-fully-optimized` which takes advantage of as many compiler flags as possible with something like:

`run-native-optimized: gcc -Ofast -funsafe-math-optimizations -march=native -flto`

karpathy:
I added one just now, thanks!

PR #19, opened 2023-07-24
Simplify Softmax

mcognetta:
Explicitly fuse two loops in softmax. It seems `-O3` doesn't catch this and fusing them results in a ~12.5% speedup on my linux machine and an ~18.5% speedup on my macbook air m2 for a 32k vocab table (for softmax only --- I don't have benchmarks for the end-to-end usage).

### DIFF
--- a/run.c
+++ b/run.c
@@ -207,15 +207,14 @@ void softmax(float* x, int size) {
             max_val = x[i];
         }
     }
-    // e^x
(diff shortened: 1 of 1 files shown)


karpathy:
I confirmed this makes a solid difference. Fascinating and thank you!

karpathy:
(I saw 181 tok/sec -> 196 tok/sec on a Linux devbox)

PR #21, opened 2023-07-24
Use embedded xoroshiro128+ prng instead of stdlib rand()

cshenton:
The C standard library `rand()` implementation is neither portable nor a particularly good prng. This PR replaces it with a small embedded xoroshiro128+ prng from https://prng.di.unimi.it/ which is public domain, well tested, and is the standard prng in many places (iirc Zig and Rust have it as their default prng).

### DIFF
--- a/run.c
+++ b/run.c
@@ -166,6 +166,42 @@ void checkpoint_init_weights(TransformerWeights *w, Config* p, FILE* f) {
 }
 
 // ----------------------------------------------------------------------------
+// pseudo-random number generation
+
+static inline unsigned int xoro_rotl(const unsigned int x, int k) {
+	return (x << k) | (x >> (32 - k));
+}
+
+static unsigned int _rng_state[4];
+
+// xoroshiro128+ PRNG https://prng.di.unimi.it/
+unsigned int xoro_rand(void) {
+	const unsigned int result = _rng_state[0] + _rng_state[3];
+	const unsigned int t = _rng_state[1] << 9;
+
+	_rng_state[2] ^= _rng_state[0];
+	_rng_state[3] ^= _rng_state[1];
+	_rng_state[1] ^= _rng_state[2];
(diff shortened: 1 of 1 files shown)


cshenton:
I've opted to use `unsigned int` instead of `uint32_t` here just to follow the existing code, but I'd be more than happy to use `stdint` and the sized type for maximum portability.

karpathy:
I am hesitant to include more lines of code here, because I'm not sure that the quality of rng makes a lot of difference in our case. The rng is only used in the softmax sampling. Is it really that bad?

cshenton:
FYI the initial state of the prng will almost certainly get overwritten in this context, I just wanted to initialise it to something in the case that it was called before `xoro_seed` was. The numbers are from the README:

> Lily felt bad achieved tok/s: 98.746993347843922

`9874 6993 3478 4392`

cshenton:
Hmm, if I'm reading the source correctly and there are only 256 weights it's sampling within, then `rand()`'s terrible random number generation might not matter too much.

Where it will matter is that `rand` only produces the same sequence of random numbers for a given seed for a given implementation. It's not portable, which means you can't necessarily share seeds of interesting inferences which seems like a shame.

karpathy:
The rand() is used to sample from a vector 32,000 probabilities. As there are 32K possible tokens. Ah ok I see, like portable!

cshenton:
`RAND_MAX` is only also only _required_ to be at least `32767` (and it actually is that value on my windows C stdlib), so the quality of the actual PRNG might start to matter there, and will break down if you exceed that scale.

aegkmq:
Perhaps [Lehmer64](https://github.com/lemire/testingRNG/blob/master/source/lehmer64.h) or [Wyrand](https://github.com/lemire/testingRNG/blob/master/source/wyrand.h) are simpler alternatives?

PR #22, opened 2023-07-24
Slightly improve tok/s

svmhdvn:
With this version and CFLAGS='-Ofast -funsafe-math-optimizations -march=native -flto', I am achieving ~97 tok/s on my old Thinkpad E495 laptop.

* replace redundant copy(...) function with standard memcpy
* reuse already existing argmax function in place of for loop
* style change: while-loop to for-loop style change to shorten loop body code visually

### DIFF
--- a/run.c
+++ b/run.c
@@ -14,6 +14,7 @@ $ ./run
 #include <stdlib.h>
 #include <time.h>
 #include <math.h>
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


karpathy:
I don't love these, but I kept the copy -> memset change and pushed a commit just now to add. ty!

PR #25, opened 2023-07-24
Add information on compiler flags

wsmoses:


karpathy:
good start i guess :)

PR #26, opened 2023-07-24
Remove duplicate if statement

BasedLukas:
Remove duplicate if statement.


karpathy:
I'll keep this because i like semantic chunks of code but ty

PR #27, opened 2023-07-24
[openmp] 1.5x inference speedup

krzysztof-jusiak:
Problem:
- clock is CPU and doesn't work properly with parallel execution.
- perf execution is matmul x weights bound.

Solution:
- use gettimeofday instead.
- utilize openmp to parallelize matmul.

Note:
- if not compiled with -fopenmp the #pragma is ignored and single execution is performed.
- there are additional env variable to setup for openmp (optinally) to setup the number of threads, scheduler etc.

Benchmarks:
```
clang -Ofast -march=native  run.c  -lm  -o run          // achieved tok/s: 340.878828
clang -Ofast -fopenmp -march=native  run.c  -lm  -o run // achieved tok/s: 524.590164
```

### DIFF
--- a/run.c
+++ b/run.c
@@ -13,6 +13,7 @@ $ ./run
 #include <time.h>
 #include <math.h>
 #include <string.h>
+#include <sys/time.h>
 
 // ----------------------------------------------------------------------------
 // Transformer and RunState structs, and related memory management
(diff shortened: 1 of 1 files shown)


Vadiml1024:
I wonder if paralleizations of:
softmax, argmax, and rmsnorm function will have a noticeable effect?

Issue #20, opened 2023-07-24
Inference speed [-Ofast]


krzysztof-jusiak:
In the spirit of the project adding additional compilation flags seems complicating things, however, `-Ofast` compilation flag seems easy to apply. Ofast is O3 + fast-math and some other optimizations (https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html).

It almost 2x the inference speed so it might be worth considering.

```
  - O3:    160t/s
  - Ofast: 307t/s
```

karpathy:
Wow, this works too. And it stacks with -O3 and -funsafe-math-optimizations.
I'm not as familiar with all of these flags, are they all present for all versions of gcc on all platforms? Are they "safe"?
Amazing.

krzysztof-jusiak:
Yeah, the Ofast is pretty standard (supported by clang and gcc for a long time).
Since Ofast is adding -fast-math that's not 100% safe as it disregard strict standards compliance, though the output for the run is the same (with O3 and Ofast), so it's pretty safe in that regards.

BTW also clang produces a bit faster inference too
`clang -Ofast run.c -o run`

```
clang-15 Ofast: 321t/s
gcc-12 0fast: 307t/s
```

It's likely that he 90% of performance is in matmul when multplying weights. One easy and portable way to tackle that without loosing much simplicity and/or portability would be to use Vector extensions (https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html).

That would allow to use SIMD in a portable way with minimal changes.

```
typedef double v4df __attribute__ ((vector_size (32)));
```

and then arithmetic operations would operate on N elements in parallel.

Additional option (less portable though) would be to use openmp for parallelism

```
void matmul(float* xout, float* x, float* w, int n, int d) {
    #pragma omp parallel for
    for (int i = 0; i < d; i++) {
        float val = 0.0f;
        for (int j = 0; j < n; j++) {
            val += w[i * n + j] * x[j];
        }
        xout[i] = val;
    }
}
```
but that would require openmp and compiling with -fopenmp but it would be quite portable and it's 1 line change.

karpathy:
I appreciate the writeup, thank you! I will take a look. I'm in favor of adding anything that will keep most of the simplicity and portability.

akx:
I wrote a small script to compile the program with different flags,
```python
import hashlib
import os

flag_sets = [
    "",
    "-O3",
    "-O3 -mtune=native -march=native",
[...]

PR #31, opened 2023-07-24
Use `clock_gettime()` instead of CPU time for benchmark

akx:
Refs https://github.com/karpathy/llama2.c/issues/20#issuecomment-1647318827

### DIFF
--- a/run.c
+++ b/run.c
@@ -370,6 +370,13 @@ int argmax(float* v, int n) {
     return max_i;
 }
 
+// get current time for benchmarking
+static double get_time() {
+    struct timespec t;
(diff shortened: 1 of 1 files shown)


karpathy:
Ty for the PR, I merged a related approach. Please have a look and re-open if you think this one is better.

e.g. GPT-4 says

*clock_gettime: This is generally a better choice than gettimeofday for measuring time intervals because it is not affected by system time changes (when using CLOCK_MONOTONIC). It provides resolution in nanoseconds, offering greater precision. However, like gettimeofday, it is a POSIX function, and may not be available on non-POSIX systems.*

PR #37, opened 2023-07-24
Have DDP ignore `freqs_cis` to avoid broadcast

awgu:
Sorry for not running DDP to catch this before merging. This PR fixes the error:
```
RuntimeError: Input tensor data type is not supported for NCCL process group: ComplexFloat
```
This solution is not very elegant, but as long as we ensure that `freqs_cis` is replicated across data parallel workers, then this is safe (which is the case since it is computed statically once and never changed thereafter). We use this private `_ddp_params_and_buffers_to_ignore` attribute to notify DDP to skip `freqs_cis` when broadcasting module states from rank 0.

I tested via:
```
torchrun --standalone --nproc_per_node=2 train.py
```
with both `compile = True` and `compile = False`.

### DIFF
--- a/train.py
+++ b/train.py
@@ -191,6 +191,10 @@ if compile:
 
 # wrap model into DDP container
 if ddp:
+    # Ignore the `freqs_cis` buffer so that DDP does not broadcast it at
+    # construction time since NCCL does not support `ComplexFloat`
+    prefix = "_orig_mod." if compile else ""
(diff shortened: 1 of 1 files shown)


karpathy:
ok, thank you for the PR. I knew there'd be some `._long_undocumented_variable_name_that_will_be_gone_next_month` to fix the problem :)

Issue #7, opened 2023-07-23
Use a BLAS lib & CLBlast for CPU & GPU speedups. [Enhancement]


trholding:
@karpathy I was thinking if you'd consider using a BLAS lib to speed up compute so that larger models may work.
If that is the case, please also have a option to compile with CLBlast (compatible drop in blas) so that compute can get offloaded to GPU via OpenCL.

https://www.netlib.org/blas/#_reference_blas_version_3_11_0
https://www.openblas.net/
https://github.com/CNugteren/CLBlast

karpathy:
Yeah... I definitely want to keep the simplicity of the repo. I'll take a look.

hchenphd:
If CPU(BLAS)&GPU(CLBlast) speedups were applied, I am very interested in benchmarking the program with different quantization model by different edge devices, such as RK3588，Nvidia Jetson Orin Nano, even Android mobile(like Qualcomm Snapdragon 7/8). 
As [karpathy](https://github.com/karpathy) said , " in this repo we focus on more narrow applications ",  I believe deploying the program on edge devices with CPU&GPU speedup can have significant value for many interesting and specific scenarios.

MostWrong:
So I tried using cblas

`
#include <cblas.h>

void accum(float *a, float *b, int size) {
    cblas_saxpy(size, 1.0f, b, 1.0f, a, 1);
}


void rmsnorm(float* o, float* x, float* weight, int size) {
    float ss = cblas_sdot(size, x, 1, x, 1);
    ss /= size;
    ss += 1e-5f;
    ss = 1.0f / sqrt(ss);

    for (int j = 0; j < size; j++) {
        o[j] = weight[j] * (ss * x[j]);
    }
}

void matmul(float* xout, float* x, float* w, int n, int d) {
    cblas_sgemv(CblasRowMajor, CblasNoTrans, d, n, 1.0, w, n, x, 1, 0.0, xout, 1);
}

`

Gives a decent speedup

trholding:
Available in a separate fork. Thanks for the discussion. Closing.

PR #45, opened 2023-07-24
Fixed OpenMP Configuration Issue for Optimal Performance

h-b:
Added a build script for Linux that detects if OpenBLAS is available and uses it in the code. On my machine it generates now 70 tokens / second instead of 60.


MostWrong:
https://github.com/karpathy/llama2.c/issues/7#issuecomment-1648132102

guess it seems fine!

karpathy:
Ok so this is the first PR that starts to get into optimizations that require ifdefs etc. Would this run faster than e.g. OpenMP? (sorry if this is a noob question)

krzysztof-jusiak:
OpenBlas is great and allows the path to the gpu easily but it also adds complexity (code side and building side). On the other hand, llama2.c has also an educational aspect going for it - how to do the pure llama inference from scratch in C  so there are trades off here. I guess it depends on how far the project wants to go. For example, llama.cpp is much faster but also it's also much more complex so there is balance to be found here, IMHO.

Anyway, interested into openblas numbers. That's what I'm getting. Do you do anything special which I'm missing?

Platform: Linux, CPU: Intel(R) Core(TM) i7-12650H

> master
```
gcc -Ofast -march=native  run.c  -lm  -o run -fopenmp
OMP_NUM_THREADS=8 ./run out/model.bin # achieved tok/s: 1137.777778                                                                                               
```

> https://github.com/karpathy/llama2.c/issues/7#issuecomment-1648132102
```
apt-get install libopenblas64-openmp
gcc -Ofast -march=native  run.c  -lm  -o run -lopenblas64
OMP_NUM_THREADS=8 ./run out/model.bin # achieved tok/s: 1007.874016                                                                                               
```

> https://github.com/karpathy/llama2.c/issues/7#issuecomment-1648132102
```
apt-get install libopenblas64-pthread
gcc -Ofast -march=native  run.c  -lm  -o run -lopenblas64
./run out/model.bin # achieved tok/s: 992.248062                                                                                                
```

BitterKanegul:
this series of optimizations is perfect to learn how things are made to go brrr from a simple implementation. 
pytorch -> simple C -> pragma omp -> blas -> gpus -> fpga_rtl(?)

krzysztof-jusiak:
That's true, thought that's really hard to capture things like that over time as the project is just getting more complex and commits history is not helping after a while. It's defo possible but it requires more work and effort but it also depends on the project goals. The way I look at it ATM is that one can always switch to llama.cpp for more. Starting from llama.cpp would be way harder than from llama.c, at least ATM. Though I have nothing against blast, cuda/metal and/or fpga (even more, I actually think perf is crucial for usability of the project) but the simplicity is very powerful concept too.

Offloading computations is part of the perf but to get more gains computation graph with optimized kernels would probably be the way to go instead of the eager execution but that's add even more complexity.

MostWrong:
I agree, part of why llama2.c is so appealing is that it is 500 lines and really easy to comprehend

h-b:
> Ok so this is the first PR that starts to get into optimizations that require ifdefs etc. Would this run faster than e.g. OpenMP? (sorry if this is a noob question)

OpenMP's default behavior of determining thread count often aligns with the number of virtual cores, which isn't optimal for our use case. On a system with 16 physical cores and 24 virtual cores, using OpenMP defaults results in speeds of just 1-2 tokens/second. By manually setting threads to the number of physical cores (even as few as 4), I achieved rates superior to the OpenBLAS matmul variant (113 vs. 83 tokens/sec). Thus, with this adjustment, the current OpenMP implementation outperforms the OpenBLAS approach.

I've noticed an overall improvement in execution times after I pulled a few minutes ago. However, the underlying issue I addressed with the thread count remains present in the repo. So, while the absolute numbers I provided above do differ now, the relative performance gains from my adjustments are still consistent and valid.

krzysztof-jusiak:
Totally agree with the observation, omp by default will take the max number of available threads which is not desired; therefore README has mentioned experimentation with the OMP_NUM_THREADS flag for better performance. I'm afraid though that `/2` is not fits all solution in this case as it's really hardware specific, not all hardware configurations will have the same. More accurate could to discover the number of available p-cores but that could be done outside the run.c, I think. The default could be changed though but I'm not sure whether it can be easily overwritten without recompiling?

h-b:
> Totally agree with the observation, omp by default will take the max number of available threads which is not desired; therefore README has mentioned experimentation with the OMP_NUM_THREADS flag for better performance. I'm afraid though that `/2` is not fits all solution in this case as it's really hardware specific, not all hardware configurations will have the same. More accurate could to discover the number of available p-cores but that could be done outside the run.c, I think. The default could be changed though but I'm not sure whether it can be easily overwritten without recompiling?

Thank you for sharing your perspective. I appreciate the insights.

To provide some context, I conducted tests on two distinct systems: one having 4 physical cores (and 8 virtual) and another with 16 physical cores (and 24 virtual). The detailed results are outlined below. A brief summary:

- On the system boasting 8 virtual cores, the performance showed slight improvement when limited to just the 4 physical cores, which aligns with the outcome of the formula.

- For the machine with 16 physical cores, the disparity in performance was striking. Using OpenMP's default setting made it run approximately 20 times slower compared to not employing OpenMP.

- It's also worth noting that on a system with a singular virtual (and physical) core, the formula produces the expected result.

Given these findings, I believe the formula should generally produce effective outcomes on conventional systems. Moreover, for those wishing to exceed the programmatically set maximum thread count, it remains feasible via the OMP_NUM_THREADS setting, as omp_get_max_threads() mirrors this count. So, for instance, if one wants to employ 24 cores despite the formula defaulting to 12, setting OMP_NUM_THREADS to 48 should suffice.

- Intel(R) Core(TM) i9-12900T
	- using OpenBLAS
		- using OpenBLAS
			- |command line|achieved tok/s|
			  |--|--|
			  |`./run out44m/model44m.bin`|83|
	- using OpenMP
		- |command line|achieved tok/s|
		  |--|--|
		  |`./run out44m/model44m.bin`|interrupted (extremely slow, around 1-2)|
		  |`OMP_NUM_THREADS=1 ./run out44m/model44m.bin`|65|
		  |`OMP_NUM_THREADS=2 ./run out44m/model44m.bin`|91|
		  |`OMP_NUM_THREADS=3 ./run out44m/model44m.bin`|102|
		  |`OMP_NUM_THREADS=4 ./run out44m/model44m.bin`|107|
		  |`OMP_NUM_THREADS=5 ./run out44m/model44m.bin`|107|
		  |`OMP_NUM_THREADS=6 ./run out44m/model44m.bin`|108|
		  |`OMP_NUM_THREADS=7 ./run out44m/model44m.bin`|105|
		  |`OMP_NUM_THREADS=8 ./run out44m/model44m.bin`|105|
		  |`OMP_NUM_THREADS=9 ./run out44m/model44m.bin`|128|
		  |`OMP_NUM_THREADS=10 ./run out44m/model44m.bin`|119|
		  |`OMP_NUM_THREADS=11 ./run out44m/model44m.bin`|111|
		  |`OMP_NUM_THREADS=12 ./run out44m/model44m.bin`|113|
		  |`OMP_NUM_THREADS=13 ./run out44m/model44m.bin`|109|
		  |`OMP_NUM_THREADS=14 ./run out44m/model44m.bin`|115|
		  |`OMP_NUM_THREADS=15 ./run out44m/model44m.bin`|109|
		  |`OMP_NUM_THREADS=16 ./run out44m/model44m.bin`|104|
		  |`OMP_NUM_THREADS=17 ./run out44m/model44m.bin`|103|
		  |`OMP_NUM_THREADS=18 ./run out44m/model44m.bin`|111|
		  |`OMP_NUM_THREADS=19 ./run out44m/model44m.bin`|106|
		  |`OMP_NUM_THREADS=20 ./run out44m/model44m.bin`|94|
		  |`OMP_NUM_THREADS=21 ./run out44m/model44m.bin`|89|
		  |`OMP_NUM_THREADS=22 ./run out44m/model44m.bin`|38|
		  |`OMP_NUM_THREADS=23 ./run out44m/model44m.bin`|19|
		  |`OMP_NUM_THREADS=24 ./run out44m/model44m.bin`|interrupted (extremely slow)|
- Intel Core i7-4810MQ @ 2.8 GHz (8 virtual cores, 4 physical cores)
	- using OpenBLAS
		- |command line|achieved tok/s|
		  |--|--|
		  |`./run out44m/model44m.bin`|53|
	- using OpenMP
		- |command line|achieved tok/s|
		  |--|--|
		  |`./run out44m/model44m.bin`|51|
		  |`OMP_NUM_THREADS=1 ./run out44m/model44m.bin`|54|
		  |`OMP_NUM_THREADS=2 ./run out44m/model44m.bin`|60|
		  |`OMP_NUM_THREADS=3 ./run out44m/model44m.bin`|66|
		  |`OMP_NUM_THREADS=4 ./run out44m/model44m.bin`|64|
		  |`OMP_NUM_THREADS=5 ./run out44m/model44m.bin`|59|
		  |`OMP_NUM_THREADS=6 ./run out44m/model44m.bin`|59|
		  |`OMP_NUM_THREADS=7 ./run out44m/model44m.bin`|59|
		  |`OMP_NUM_THREADS=8 ./run out44m/model44m.bin`|51|

Please note that while the absolute numbers I provided above do differ with the latest commits, the relative performance gains from my adjustments are still consistent and valid.

h-b:
> [#7 (comment)](https://github.com/karpathy/llama2.c/issues/7#issuecomment-1648132102)
> 
> guess it seems fine!

Thank you for taking the time to review my code and for your feedback. Also, thanks for directing my attention to the existing discussion and the experimentation with cblas.

I have evaluated the `accum` and `rmsnorm` functions you've shared. Based on my tests, I couldn't observe a noticeable speedup compared to the original implementation. As for `matmul`, my observation aligns with yours when using `cblas_sgemv` from OpenBLAS. However, when properly configured with OpenMP (as proposed in my updated pull request), the performance exceeded that of the BLAS variant.

karpathy:
Sounds like we're not seeing improvements from OpenBLAS (somewhat surprisingly)... so I'll close this PR for now. Feel free to re-open if disagreements. If OpenBLAS was faster I'd be willing to consider a bit more complex matmul, because the performance of the system is so dependent on this one element.

h-b:
> Sounds like we're not seeing improvements from OpenBLAS (somewhat surprisingly)... so I'll close this PR for now. Feel free to re-open if disagreements. If OpenBLAS was faster I'd be willing to consider a bit more complex matmul, because the performance of the system is so dependent on this one element.

I appreciate your feedback. However, I'd like to clarify a crucial aspect of the changes proposed in this PR.

Thank you for reviewing the PR. I understand your perspective regarding the inclusion of OpenBLAS. However, I'd like to clarify that the key improvement in this PR isn't specifically tied to OpenBLAS. The primary enhancement is the OpenMP configuration fix, which addresses a critical performance issue.

On systems with many cores, the software was underperforming drastically in the default OpenMP configuration. To give a specific example: on a system with 16 physical cores, the software was achieving only 1-2 tok/s with the default configuration. By optimizing the number of threads OpenMP uses, performance increased dramatically, achieving up to 128 tok/s when approximately half of the virtual cores were utilized. This optimization ensures that the software performs efficiently across multi-core systems.

Given the importance of this fix, I'd like to kindly ask if you'd consider reopening this PR to take another look. Unfortunately, it seems I don't have the option to reopen it myself. I renamed the PR to better reflect the issue it addresses.

Thank you for your understanding and consideration.

karpathy:
I think any application should be able to tune the number of threads to their case. This is a hyperparameter people should sweep per application. I'm not sure that it makes sense to assume or hardcode its configuration in code?

h-b:
First and foremost, I'd like to address the user experience for those trying out the application on average desktop systems. When running on a common processor like the Intel(R) Core(TM) i9-12900T, the application experiences a drastic performance reduction by roughly a factor of 100 when using OpenMP's default configuration. This magnitude of a slowdown might lead many users to perceive the application as malfunctioning, especially if they're unaware of the underlying configuration intricacies.

I absolutely agree that users should be able to configure the number of threads based on their specific use-case. However, the default setting should provide a reasonable baseline performance. In the current scenario, OpenMP's default configuration doesn't yield that on systems with more than 8 virtual cores. Even on an 8-core system (with 4 physical and 4 virtual cores), utilizing all 8 cores results in about a 20% decrease in performance compared to restricting OpenMP threads to 4.

The intricacies of OpenMP's thread management can indeed be non-trivial. OpenMP does make its own decisions on the number of threads to use, and while OMP_NUM_THREADS sets an upper limit, the actual behavior is determined dynamically by OpenMP's internal strategies. These strategies, while generally effective, are not foolproof and can underperform in specific cases, as observed with this application.

While it would be ideal to limit the number of threads to physical cores, querying this number reliably at runtime can be challenging. And although OpenMP provides mechanisms like omp_set_dynamic(1) to influence its strategy, this too doesn't remedy the observed performance issue in our case.

Lastly, I'd like to highlight that in its default configuration, OpenMP underperforms compared to OpenBLAS, even on laptop systems. This means the initial observation about not seeing improvements from OpenBLAS isn't entirely accurate.

In summary, while the flexibility to set the number of threads is essential, the default configuration should be robust enough to provide satisfactory performance on a broad range of systems. My pull request aims to bridge this gap, ensuring that users get a reliable experience out of the box.

Thank you for your time and consideration.

PR #47, opened 2023-07-24
Quick fix for MSVC WIN32 compatibility

richinseattle:
Use GetTickCount() on Windows instead of gettimeofday()


karpathy:
I hate ifdefs. Code currently on master doesn't work on Windows?

richinseattle:
Windows doesn't have gettimeofday() even with <sys/timeb.h>. could use rdtsc for compatibility instead.

PR #48, opened 2023-07-24
MSVC Compatibility fix for timer

richinseattle:
use clock() instead of gettimeofday() for cross-platform compatibility

### DIFF
--- a/run.c
+++ b/run.c
@@ -377,7 +377,9 @@ int argmax(float* v, int n) {
 // ----------------------------------------------------------------------------
 
 long time_in_ms() {
-  return (1000 * clock()) / CLOCKS_PER_SEC;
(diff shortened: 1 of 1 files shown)


karpathy:
I just changed it the other way around. clock() measures CPU time, doesn't report correct time when using OpenMP

richinseattle:
Ah, on Windows clock() measures wall time but cpu time on posix.. I'm not sure how to avoid an ifdef to reliably get wall time cross platform. My original patch with GetTickCount() is the closest replacement for gettimeofday()

richinseattle:
I've tested timespec_get() on Windows and Linux with openmp and it appears to work as expected for wall clock time

karpathy:
ok let's try.

PR #52, opened 2023-07-25
Add CMake build

Maknee:
* Added CMakeLists.txt (Having LLAMA_C_OFAST and LLAMA_C_NATIVE options)
* Added a section to `README.md` about building with CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,32 @@
+cmake_minimum_required(VERSION 3.11)
+project(llama.c)
+
+set(CMAKE_C_STANDARD 11)
+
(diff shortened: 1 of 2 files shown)


sidsarasvati:
I like the simplicity of single run.c file and lovely Makefile so my vote is to not add cmake unless it much needed
My $0.02

karpathy:
agree but ty

karpathy:
(I'm not as familiar with build systems in C. What are the advantages of cmake over make here? I've had a lot of scaring experiences with cmake in the past, so very suspicious of it. Please re-open the issue if you'd like to contest this)

PR #57, opened 2023-07-25
Update Makefile

tmc:
drops unnecessary .PHONY


karpathy:
? it's not unnecessary, it lets you run the same command and recompile each time.

PR #60, opened 2023-07-25
Fixed time_in_ms() compile time error (termux and neoterm)

emma-eva:
clang version 16.0.4


karpathy:
thank you. i believe this one doesn't work on windows, though? also not sure if it works correctly with omp.

PR #68, opened 2023-07-25
Load weights at startup

xefoci7612:
Weights are loaded automatically during the first iteration, but this can take few seconds, just after 's' is printed and before the first token is shown. Instead preload weights at mapping time (before 's' is printed) and give some feedback to the user.


xefoci7612:
BTW considering that _all_ the weights are loaded in RAM anyway at first iteration, it is not so clear the point of memory mapping them instead of loading and memory accessing them directly.

Memory mapping has a point if the weights file is bigger than available RAM, but in that case swapping would be very slow anyhow.

karpathy:
ty but i think ok as is (?)

PR #74, opened 2023-07-25
fix bug of timings

bjmsong:


karpathy:
omg...

karpathy:
I didn't want to include your gitignore so i pushed a commit just now for fix, ty

PR #77, opened 2023-07-25
Update README.md: formate output samples

madroidmaq:
| Before | After |
| ------ | ----- |
|     image  |    image  |


karpathy:
neat ty, better :)

PR #90, opened 2023-07-26
use os.path.join() instead of joining strings for params.json

yamada321:
Hi, 

Proposing a very small modification. In the current HEAD's `export_meta_llama_bin.py`, user needs to supply path/to/llama/ (with the trailing slash) because the path to params.json is obtained by string concatenation. The PR changes it to os.path.join(), and also checks the file with os.path.isfile() & prompts an error if false. 

Thank you.


karpathy:
I fixed it a bit different ty

PR #93, opened 2023-07-26
Support for different data types (float16, float32)

krzysztof-jusiak:
Problem:
- Only float32 is currently supported which requires 2x the memory and it slower to execute as inference is memory bound.

Solution:
- Add ability to export model to float16.
- Add support to inference with float16.

Note:
- It's quite hard to do in generically in pure C (without templates) so to avoid adding too much complexity compilation option has been chosen. Ideally that would be a run-time pick based on the value stored in the config but that requires additional complexity which I wanted to avoid but that can be still explored with a proper solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,10 @@ This still runs at interactive rates and samples more coherent and diverse stori
 As the neural net architecture is identical, we can also inference the Llama 2 models released by Meta. Sadly there is a bit of friction here due to licensing (I can't directly upload the checkpoints, I think). So Step 1, get the Llama 2 checkpoints by following the [Meta instructions](https://github.com/facebookresearch/llama). Once we have those checkpoints, we have to convert them into the llama2.c format. For this we use the `export_meta_llama_bin.py` file, e.g. for 7B model:
 
 ```bash
-python export_meta_llama_bin.py path/to/llama/model/7B llama2_7b.bin
+python export_meta_llama_bin.py path/to/llama/model/7B llama2_7b.bin float32
 ```
 
-The export will take ~10 minutes or so and generate a 26GB file (the weights of the 7B model in float32) called `llama2_7b.bin` in the current directory. It has been [reported](https://github.com/karpathy/llama2.c/pull/85) that despite efforts, the 13B export currently doesn't work for unknown reaons (accepting PRs for fix). We can run the model as normal:
+The export will take ~1 minute or so and generate a 26GB file (the weights of the 7B model in float32) called `llama2_7b.bin` in the current directory. It has been [reported](https://github.com/karpathy/llama2.c/pull/85) that despite efforts, the 13B export currently doesn't work for unknown reaons (accepting PRs for fix). We can run the model as normal:
 
 ```bash
 ./run llama2_7b.bin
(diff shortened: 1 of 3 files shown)


vgoklani:
@krzysztof-jusiak Hey there, what about approach would be used for BF16?

thanks!

axrwl:
For clang, `-DDTYPE=_Float16` will work.

xefoci7612:
@krzysztof-jusiak: Hi, do you think it makes sense to just move from float32 to float16? It seems float16 is always faster than float32.

Using a single data type would be simpler both in code and in numbers of modelxxx.bin (already 3 + llama 7B and counting...)

README.md
```diff
@@ -46,10 +46,10 @@ This still runs at interactive rates and samples more coherent and diverse stori
...
 
 ```bash
-python export_meta_llama_bin.py path/to/llama/model/7B llama2_7b.bin
+python export_meta_llama_bin.py path/to/llama/model/7B llama2_7b.bin float32
 ```
 
-The export will take ~10 minutes or so and generate a 26GB file (the weights of the 7B model in float32) called `llama2_7b.bin` in the current directory. It has been [reported](https://github.com/karpathy/llama2.c/pull/85) that despite efforts, the 13B export currently doesn't work for unknown reaons (accepting PRs for fix). We can run the model as normal:
+The export will take ~1 minute or so and generate a 26GB file (the weights of the 7B model in float32) called `llama2_7b.bin` in the current directory. It has been [reported](https://github.com/karpathy/llama2.c/pull/85) that despite efforts, the 13B export currently doesn't work for unknown reaons (accepting PRs for fix). We can run the model as normal:
```
kroggen, line 52: reaons -> reasons

krzysztof-jusiak:
@xefoci7612 I think there is a need for different data type as they provide different accuracy and differ in sizes, etc. Ideally we would operate on max precision floating point only but that's not possible due to performance/size limitations. However, smaller models or better hardware allow to use higher precision versions. Since inference is memory bound for faster speeds usually quantization is also applied, which compress the model even more to for example 4 bits but that comes with accuracy trades off, which I don't think can be made by default for everyone.

kroggen:
@xefoci7612 

No, float16 is not always faster than float32. If the processor does not natively support FP16 arithmetic, then it will be emulated in software

https://stackoverflow.com/questions/56697332/float16-is-much-slower-than-float32-and-float64-in-numpy

karpathy:
will def take a look here. one thing to be careful with is that if you want to inference in fp16 you _must_ train in fp16 (with gradient scalers), and not in fp32 or bf16. Otherwise the range of activations can overflow.

karpathy:
@krzysztof-jusiak what are the benefits of fp16? 

- clearly checkpoints get half as small
- how much faster is fp16? e.g. on your computer, or on a MacBook Air M1 or so

also my understanding is this would invalidate all our previous checkpoints because they don't contain dtype in the config. Which is on me for having chosen a dumb serialization :D

krzysztof-jusiak:
Thanks for the hint about the training, it's a really valid point.

Regarding the benefits of fp16, it's mainly performance as the weights size is smaller and since the computations are memory bound the speedup should be noticeable.

Saying that, I've not noticed a huge difference between fp16 and fp32 on my machine
fp16: 2.3tok/s
fp32: 2.0tok/s

which doesn't add up with my previous experiments with other llms, I'd expect a bit bigger difference.
I'm verifying whether the code is at memory bound stage or not yet. Maybe it's still cpu bound which would explain why the speedup isn't bigger.

axrwl:
> also my understanding is this would invalidate all our previous checkpoints because they don't contain dtype in the config.
Backwards compatibility

axrwl:
> will def take a look here. one thing to be careful with is that if you want to inference in fp16 you _must_ train in fp16 (with gradient scalers), and not in fp32 or bf16. Otherwise the range of activations can overflow.

Ba

axrwl:
> also my understanding is this would invalidate all our previous checkpoints because they don't contain dtype in the config

Backwards compatibility can be maintained by assuming that a missing dtype implies fp32.

run.c
```diff
@@ -17,6 +17,10 @@ Then run with:
 #include <fcntl.h>
 #include <sys/mman.h>
 
+#ifndef DTYPE
```
ozabluda, line 20: It's better to typedef dtype rather than spreading DTYPE macro all over.
```suggestion
#ifdef DTYPE
typedef DTYPE dtype;
else
typedef float dtype;
#else

```

karpathy:
Ok I took a look but I think the required changes are a little bit too yucky, and we're not seeing strong evidence of much better results. I do like that the files would have been smaller by ~half... Anyway I'm leaning to not include this atm but ty. If someone can demonstrate a solid higher throughput I think it's worth revisiting.

Issue #91, opened 2023-07-26
Can't compile run.c on windows platform via MSVC/cl.exe


tairov:
Seems that on Windows it's not possible to compile run.c via MSVC compilers

```
cl.exe run.c
cl.exe run.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.35.32217.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

run.c
run.c(16): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
mingw32-make: *** [makefile:42: windowscl] Error 2
Error: Process completed with exit code 1.
```

System Info:
```
Microsoft Windows Server 2022
  10.0.203[4](https://github.com/tairov/llama2.c/actions/runs/5663591697/job/15345560107#step:1:4)8

Image: windows-2022
  Version: 20230716.1.0
  Included Software: https://github.com/actions/runner-images/blob/win22/20230716.1/images/win/Windows2022-Readme.md
  Image Release: [https://github.com/actions/runner-](https://github.com/actions/runner-images/releases/tag/win22%2F20230716.1)
```

MSBuild: 
```
C:\ProgramData\Chocolatey\bin\vswhere.exe -products * -requires Microsoft.Component.MSBuild -property installationPath -latest
C:\Program Files\Microsoft Visual Studio\2022\Enterprise
```

msvc-dev-cmd:
```
Found with vswhere: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat
```

See GH Actions execution (Build step, errors ignored) https://github.com/tairov/llama2.c/actions/runs/5663591697/job/15345560107

Also tried on Windows 10 virtual machine with installed VS Build Tools including Windows 11 SDK, similar error:

```
cl run.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.36.32537 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

run.c
run.c(16): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
```

PR with CI job to test builds on different OS is here https://github.com/karpathy/llama2.c/pull/86

xunboo:
windows build is passing on https://github.com/xunboo/llama2win

tairov:
@xunboo do you have any idea how to adapt `run.c` in current repo? It must better for maintenance and CI workflows

xunboo:
[...]

PR #96, opened 2023-07-26
Add Windows mmap & timing compatibility

tairov:
Compilation on Windows platform was broken after merging mmap PR #50 
This PR add compatibility as it was suggested in this issue #91 

CI: https://github.com/tairov/llama2.c/actions/runs/5664313188/job/15347372920

### DIFF
--- a/mmap-windows.c
+++ b/mmap-windows.c
@@ -1,0 +1,100 @@
+/* mmap() replacement for Windows
+ *
+ * Author: Mike Frysinger <vapier@gentoo.org>
+ * Placed into the public domain
+ */
+
+/* References:
+ * CreateFileMapping: http://msdn.microsoft.com/en-us/library/aa366537(VS.85).aspx
+ * CloseHandle:       http://msdn.microsoft.com/en-us/library/ms724211(VS.85).aspx
+ * MapViewOfFile:     http://msdn.microsoft.com/en-us/library/aa366761(VS.85).aspx
+ * UnmapViewOfFile:   http://msdn.microsoft.com/en-us/library/aa366882(VS.85).aspx
+ */
+
+#include <io.h>
+#include <windows.h>
+#include <sys/types.h>
+
+#define PROT_READ     0x1
+#define PROT_WRITE    0x2
+/* This flag is only available in WinXP+ */
+#ifdef FILE_MAP_EXECUTE
+#define PROT_EXEC     0x4
+#else
+#define PROT_EXEC        0x0
+#define FILE_MAP_EXECUTE 0
+#endif
+
+#define MAP_SHARED    0x01
+#define MAP_PRIVATE   0x02
+#define MAP_ANONYMOUS 0x20
+#define MAP_ANON      MAP_ANONYMOUS
+#define MAP_FAILED    ((void *) -1)
+
+#ifdef __USE_FILE_OFFSET64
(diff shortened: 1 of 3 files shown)


karpathy:
@tairov this fixes both timing and mmap for windows, right? can you say a bit more about where this code comes from? i expect there are no easy alternatives that don't require additional files to be added to the project?

xunboo:
I got the mmap file from https://github.com/m-labs/uclibc-lm32/blob/master/utils/mmap-windows.c
However, I suggest use .h instead of .c for header name

tairov:
@karpathy yes, it fixes issue with timing and mmap.
As @xunboo mentioned `mmap-windows.c` is from  [uclibc-lm32 repo](https://github.com/m-labs/uclibc-lm32/blob/master/utils/mmap-windows.c) with [GPL v2.1 License](https://github.com/m-labs/uclibc-lm32/blob/master/COPYING.LIB)

I would say there is no simple solution that can keep `run.c` simple without additional file with encapsulated windows logic `mmap/CreateFileMappingA`.
I think it's much more easier to maintain code by having `#if #else` pragma check to make it compatible with Windows.
`llama.cpp` also uses `#if defined(_WIN32)` checks but directly from [source](https://github.com/ggerganov/llama.cpp/blob/32c54116318929c90fd7ae814cf9b5232cd44c36/llama-util.h#L212-L211) 

Considering that the beauty of this repository is its simplicity, and also considering the requests of the community to [keep things simple](#104), I see this solution with an additional file as more reasonable for supporting the Windows platform.

tairov:
Simplified `time_in_ms`, got rid of `#ifdefs`. 
Moved `mmap & clock_gettime` ported versions to `win.h` that is included in a single place in `run.c` with apparently inevitable `#if defined` pragma check.

karpathy:
@tairov IANAL etc but isn't GPL code not compatible with this repo, if it is under MIT?

tairov:
Generally yes, seems that GPL is not compatible with MIT.
Since we're using this file only https://github.com/m-labs/uclibc-lm32/blob/master/utils/mmap-windows.c , it says 
```
 * Placed into the public domain
```

ChatGPT meaning on this 😄  : 

> Public domain code refers to software that is not protected by copyright. It is available for anyone to use for any purpose, without needing to obtain permission or pay licensing fees.

richinseattle:
I've fixed these issues in my fork yesterday, including the fix for timer, mmap (using the mman-win32), and also today fixed the openmp pragmas to work with msvc. Compiles with msvc or cross-compile with mingw32. Makefile target and a .bat added for building.
https://github.com/richinseattle/llama2.c

karpathy:
@tairov in this version how does the windows-specific version of `clock_gettime` end up getting called? There's no ifdef needed?

richinseattle:
The function implementation is hidden in the win.h header so the only ifdef is in the includes. Not a bad way to do it.

tairov:
@karpathy I simplified `run.c` and `clock_gettime` alternative that is also moved to [win.h:#112](https://github.com/karpathy/llama2.c/pull/96/files#diff-72db5fdcbf1e9ca253896088a1a03b3704e927562a67f9ff8d4378e367c4fec7R112)
Overall `win.h` now contains 112 lines and solves both existing issues with compilations on windows

karpathy:
Merged the other fork but ty @tairov for all the help here!

PR #97, opened 2023-07-26
Improved Serialization and Suggested Changes for File Header in Export Script

Foundation42:
This pull request provides a significant performance improvement to the model export process by introducing more efficient serialization:

```python
# This is purely for the float32 format of the current llama.c
# and should be adjusted once support for alternate types are properly implemented
def serialize(t):
    d = t.detach().cpu().float().numpy()
    b = d.flatten().tobytes()
    f.write(b)
```

In addition to the performance improvements, this pull request also suggests a method for enhancing the file header. It introduces a **header_magic_version** integer which serves three purposes:

1. Identifying the file type via a magic byte (in this case 0x42)
2. Providing versioning information in the next three bytes (0xMajor, 0xMinor, 0xPoint)
3. Padding the header for cache alignment (typically 32 bytes)

For now, we have omitted the emission of the header_magic_version, and only left the comment, but it is recommended to consider this for future updates.

```python
    # header magic version integer added for three reasons
    # (this PR does not address the alignment or versioning problems with the current header
    # although at some point this should be re-enabled)
    #
    # the format of the header_magic_version is three-fold
    #
    # 1. to provide identification of the file-type via a magic byte (in this case 0x42)
    # 2. to provide versioning information in the next three bytes
    #    0xMajor 0xMinor 0xPoint
    # 3. to pad the header for cache alignment (typically 32 bytes)
    # for now we will skip emitting the header_magic_version
    #
    # header_magic_version = 0x42000000
    # header = struct.pack('iiiiiiii', header_magic_version, p['dim'], hidden_dim, n_layers, p['n_heads'], 
    #                                n_kv_heads, -p['vocab_size'], p['max_seq_len'])
    #
    header = struct.pack('iiiiiii', p['dim'], hidden_dim, n_layers, p['n_heads'], 
                                    n_kv_heads, -p['vocab_size'], p['max_seq_len'])
```

This approach would introduce more flexibility and better alignment with the cache, improving the overall performance of the system, as well as providing a much needed versioning scheme for the file format.

See prior PR discussion for further info
#94

### DIFF
--- a/export_meta_llama_bin.py
+++ b/export_meta_llama_bin.py
@@ -35,22 +35,23 @@ def export(checkpoint, params, filepath='model.bin'):
     p['vocab_size'] = 32000
     n_layers = p['n_layers']
     n_kv_heads = p.get('n_kv_heads', p['n_heads'])
-    # header magic version integer added for two reasons
-    # 1) so that we can version the header
-    # 2) so that the struct maintains strict cache alignment
-    #    which is necessary so that the weights that follow the header are also cache aligned
 
-    # this PR does not address the alignment or versioning problems with the current header
-    # at some point this should be re-enabled
-    # header_magic_version = 0x42000000
+    # header magic version integer added for three reasons
+    # (this PR does not address the alignment or versioning problems with the current header
+    # although at some point this should be re-enabled)
+    #
     # the format of the header_magic_version is three-fold
+    #
     # 1. to provide identification of the file-type via a magic byte (in this case 0x42)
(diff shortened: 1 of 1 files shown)


sochi-bae:
good

ozabluda:
>Padding the header for cache alignment (typically 32 bytes)
On x86 cache alignment is 64 bytes since Pentium 4. On ARM it's also 64 and on Apple M* it's 128. I don't know any still relevant CPUs with cache line of 32.

PR #101, opened 2023-07-26
Mask out initial weights load time

xefoci7612:
I don't expect this to be committed, but....

Initial load from disk of weights takes a sensible time compared to the total run. In my case around 50% of time is spent on the first token alone.

This time is not related to run.c code by itself, it is just the unavoidable time to load from disk the weights at startup, so maybe it makes no sense to report when profiling for different algorithms. It is more an init time than the speed of generating tokens.

### DIFF
--- a/run.c
+++ b/run.c
@@ -446,7 +446,7 @@ int main(int argc, char *argv[]) {
     malloc_run_state(&state, &config);
     
     // the current position we are in
-    long start = time_in_ms();
+    long start = 0; // init after the initial load of weights
     int next;
     int token = 1; // 1 = BOS token in Llama-2 sentencepiece
     int pos = 0;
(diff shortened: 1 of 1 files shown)


kroggen:
I was not understanding this, because the code to "load" the weights comes before...

Then I realized that mmap loads data into memory on the first use

xefoci7612:
> I was not understanding this, because the code to "load" the weights comes before...
> 
> Then I realized that mmap loads data into memory on the first use

Actually mmap() just maps the file into (virtual) memory. It is the actual referencing to the data that indirectly 'loads'  the weights by means of the page fault mechanism. This is done at the first call of transformer() function that uses **all** the weights in the file and so forces the 'loading' of the whole file from disk into memory.

karpathy:
I just pushed a fix ty

PR #103, opened 2023-07-26
Refactor freqs_cis into freqs_cos and freqs_sin

ai-doge:
Refactor freqs_cis into freqs_cos and freqs_sin
remove complex64 for ONNX export compatibility

### DIFF
--- a/export_meta_llama_bin.py
+++ b/export_meta_llama_bin.py
@@ -55,10 +55,10 @@ def export(p, state_dict, filepath='model.bin'):
 
     # final rmsnorm
     serialize('norm.weight')
-    # freqs_cis
-    freqs_cis = precompute_freqs_cis(p['dim'] // p['n_heads'], p['max_seq_len'] * 2)
-    state_dict['freqs_cis.real'] = freqs_cis.real[:p['max_seq_len']]
-    state_dict['freqs_cis.imag'] = freqs_cis.imag[:p['max_seq_len']]
+    # freqs_cos, freqs_sin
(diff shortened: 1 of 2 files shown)


vgoklani:
Can the updated model be compiled with the torch.compile() function?

tcapelle:
The `torch.polar` solution was so elegant 😭

dongwangdw:
> The `torch.polar` solution was so elegant 😭

yes, agree. But currently triton does not support complex value which is the backend of pytorch.

dongwangdw (commented):
```
        serialize(self.freqs_cis.real[:p.max_seq_len])
        serialize(self.freqs_cis.imag[:p.max_seq_len])
```
With your change, this freqs_cis is not registered here
These two lines should be fix too. As I tried to compile and it failed at this.

ai-doge:
> @ai-doge
> 
> ```
>         serialize(self.freqs_cis.real[:p.max_seq_len])
>         serialize(self.freqs_cis.imag[:p.max_seq_len])
> ```
> 
> With your change, this freqs_cis is not registered here These two lines should be fix too. As I tried to compile and it failed at this.

Thank you for pointing out the error, the issue has been fixed.

karpathy:
Thank you for this PR. The official Meta implementations of Llama2 should have imo used an implementation along these lines. Complex64 does not have enough existing support on ONNX, torch.compile, and is unnecessarily confusing to (I expect most) people not familiar with complex numbers.

vgoklani:
Hey @ai-doge @dongwangdw Are you able to use torch.compile() on the updated code? I get a stack-trace, where it hits the config.cache_size_limit (64)... thanks!

PR #105, opened 2023-07-26
llama2.rs

gaxler:
I've ported Karpathy's llama2.c to Rust! 🦀 It's just as minimalistic as the original C code.
This is a quick-and-dirty first attempt.

Why? Because it was FUN! Plus, I'm curious to see how the C and Rust versions will evolve differently.

Performance is on par with C (w/o `-Ofast`). Benchmarked on my dev VM.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+llama2-rs/target/
--- a/README.md
+++ b/README.md
@@ -1,186 +1,33 @@
-## llama2.c
 
-<img src="assets/llama_cute.jpg" width="300" height="300">
+## llama2.rs 🦀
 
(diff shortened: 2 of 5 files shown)


karpathy:
Very cool! Probably should be be a separate repo? 🤔

antonkesy:
> Very cool! Probably should be be a separate repo? thinking

agree! 
please put it in its own repository

gaxler:
> Very cool! Probably should be be a separate repo? 🤔

Thanks!
Hope it can keep up with the high pace of this repo.

If anyone reading is interested in contributing:
https://github.com/gaxler/llama2.rs

PR #111, opened 2023-07-26
Fix 13B

python273:
Screenshot 2023-07-26 at 16 43 26

### DIFF
--- a/run.c
+++ b/run.c
@@ -132,11 +132,11 @@ void checkpoint_init_weights(TransformerWeights *w, Config* p, float* f, int sha
     w->rms_ffn_weight = ptr;
     ptr += p->n_layers * p->dim;
     w->w1 = ptr;
-    ptr += p->n_layers * p->dim * p->hidden_dim;
+    ptr += (unsigned int)p->n_layers * p->dim * p->hidden_dim;
     w->w2 = ptr;
-    ptr += p->n_layers * p->hidden_dim * p->dim;
+    ptr += (unsigned int)p->n_layers * p->hidden_dim * p->dim;
     w->w3 = ptr;
-    ptr += p->n_layers * p->dim * p->hidden_dim;
+    ptr += (unsigned int)p->n_layers * p->dim * p->hidden_dim;
     w->rms_final_weight = ptr;
     ptr += p->dim;
     w->freq_cis_real = ptr;
@@ -222,7 +222,7 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
     float* freq_cis_imag_row = w->freq_cis_imag + pos * head_size / 2;
 
     // forward all the layers
-    for(int l = 0; l < p->n_layers; l++) {
+    for(unsigned int l = 0; l < (unsigned int)p->n_layers; l++) {
 
         // attention rmsnorm
         rmsnorm(s->xb, x, w->rms_att_weight + l*dim, dim);


xefoci7612:
What about setting the correct type directly in the definition and avoid a bunch casts?

```C
typedef struct {
    size_t dim; // transformer dimension
    size_t hidden_dim; // for ffn layers
    size_t n_layers; // number of layers
    size_t n_heads; // number of query heads
    size_t n_kv_heads; // number of key/value heads (can be < query heads because of multiquery)
    size_t vocab_size; // vocabulary size, usually 256 (byte-level)
    size_t seq_len; // max sequence length
} Config;
```

size_t is also more idiomatic than int or better unsigned in when dealing with pointer arithmetic

python273:
It's loaded from a model file, so would need a conversion function. vocab_size also can be negative (used as a flag)

xefoci7612:
> It's loaded from a model file, so would need a conversion function. vocab_size also can be negative (used as a flag)

I see thanks

`if(fread(&config, sizeof(Config), 1, file) != 1) { return 1; }
`

In this case it works by accident, it assumes _int_ has a fixed size, but _int_ can have different sizes according to the platform. At least it should be specified the size of Config struct fields as int32_t

typedef struct {
    int32_t dim; // transformer dimension
    int32_t hidden_dim; // for ffn layers
    int32_t n_layers; // number of layers
    int32_t n_heads; // number of query heads
    int32_t n_kv_heads; // number of key/value heads (can be < query heads because of multiquery)
    int32_t vocab_size; // vocabulary size, usually 256 (byte-level)
    int32_t seq_len; // max sequence length
} Config;

I have verified the size remains the same (28 bytes).

...but using signed integer instead of an unsigned size can yield to bugs with big files, as your patch illustrates.

python273:
hmm, lots of warnings though. I'm not a C dev, so not sure what's a good practice here 🤔 
```
run.c:260:27: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
        for (int h = 0; h < p->n_heads; h++) {
                        ~ ^ ~~~~~~~~~~
run.c:221:22: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
    for(int l = 0; l < p->n_layers; l++) {
                   ~ ^ ~~~~~~~~~~~
run.c:232:27: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
        for (int h = 0; h < p->n_heads; h++) {
                        ~ ^ ~~~~~~~~~~
run.c:307:27: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
        for (int i = 0; i < hidden_dim; i++) {
                        ~ ^ ~~~~~~~~~~
run.c:312:27: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
        for (int i = 0; i < hidden_dim; i++) {
                        ~ ^ ~~~~~~~~~~
run.c:424:29: warning: comparison of integers of different signs: 'int' and 'uint32_t' (aka 'unsigned int') [-Wsign-compare]
    if (steps <= 0 || steps > config.seq_len) { steps = config.seq_len; }
                      ~~~~~ ^ ~~~~~~~~~~~~~~
6 warnings generated.
```

python273:
Yeah, not sure if these warnings are a problem in practice. `-Wsign-compare` I fixed is under `-Wextra`, so dunno, probably can revert back for now?

xefoci7612:
@python273 compiling with

`gcc -Wall -Wextra -Ofast -fopenmp -march=native run.c  -lm  -o run`

I don't get warnings. Only with 

`gcc -Wsign-conversion -Ofast -fopenmp -march=native run.c  -lm  -o run`

I have a lot of warnings.

I'm not able to test the 7B model, the smaller ones still work with your patch and if for you now the 7B is also fixed, I'd suggest this patch is good.

Auditing the types to be consistent across the source is not a 5 minute job because choices should be made, that will remain for the future, so maybe it can be deferred to a future specific PR IMHO.

python273:
@karpathy can be merged I guess

For cleaner diff, `-Wextra` gives warnings (specifically `-Wsign-compare` mentioned above, which shouldn't be a problem in practice), I'll let someone more experienced with C figure out how to handle it better

raimondasl:
I agree with @xefoci7612 that the changes should be merged as is. 
I also agree with @xefoci7612 that ideally the types should be cleaned up and that it is not a 5 minute job. It will touch a lot of lines in the file and make some parts a bit uglier.
So @karpathy - plz comment or open an issue for the types cleanup. I can do it, but I'd rather get a go-ahead for such changes. Thx.

karpathy:
Does this change invalidate the previous .bin checkpoints, and require me to re-export and re-upload all checkpoints?

xefoci7612:
@karpathy this patch is compatible with current modelxx.bin, but I didn't test the export tool from LLama to *.bin

Eventually we can just leave the types as they are, and only change int -> int32_t in struct Config to unambiguously match the 'iiiii' in struct.pack()

This will overflow for 7B, but maybe is possible to just change the offset type to size_t (the correct type to use) and fix the 7B case, i.e. reduce the fix to the minimum to make it work.

We can defer the type rework to a following patch, eventually when the *.bin file structure is consolidated and fixed.

python273:
Existing model files still work. If you want to change some fields to signed int later, you still can, I just switched `struct.pack` to unsigned to match C code

ozabluda:
I can help with types rework

karpathy:
@python273 is it not the case that the python files can be left alone? it's only in the run.c file where the issue is, because the datatypes are ints and they overflow in the pointer arithmetics?

xefoci7612:
Here is my attempt at it #153

kroggen:
@karpathy 

Yes. There is no need to change the python scripts. The overflow problem is on `run.c`

The proof is the model running on tinygrad

kroggen:
I sent a fix with minimal changes on #154

python273:
@karpathy sure, `i` vs `I` doesn't matter with small numbers, but if you want to add sign to other field (like with vocab_size), it would be easier to remember you need to update both places as the types are matching in Py and C, and Py at least would raise an exception 😅 I can revert if you want  ¯\\\_(ツ)\_/¯

@kroggen some offsets are calculated for each layer https://github.com/karpathy/llama2.c/pull/111/commits/dc3c08d5055c3ccbfaabba89db807b88ebfe551e (see matmuls)

kroggen:
I updated the `transformer()` function based on that. Thank you @python273 !

python273:
@karpathy  It fixes 13B, but I think 70B would fail to load (needs size_t)

raimondasl:
@karpathy : I prefer #153, since it also fixes 
```
-       float* weights_ptr = data + sizeof(Config)/sizeof(float);
+       float* weights_ptr = data + sizeof(Config)/sizeof(size_t);
```
Your call though. :)

karpathy:
#153 looks like has a lot going on. I'm really not a fan of this size_t stuff. I like int and float and such. Super simple. What the hell is size_t, it looks ugly.

karpathy:
[This article](https://www.embedded.com/why-size_t-matters/) clearly states that 

>Classic C (the early dialect of C described by Brian Kernighan and Dennis Ritchie in The C Programming Language , Prentice-Hall, 1978) didn't provide size_t .

i.e. the original (and perfect) C did not include this abomination.

raimondasl:
@karpathy 
> #153 looks like has a lot going on. I'm really not a fan of this size_t stuff. I like int and float and such. Super simple. What the hell is size_t, it looks ugly.

OK. :) I'd say merge this one then and we'll deal with any additional issues when/if we need to support 70B model.

karpathy:
I'm not going to merge any mention of size_t into llama2.c. Are we able to make the smallest surgical change that fixes 13B load using a simple long / unsigned long instead of ints?

python273:
@karpathy is it surgical enough? 😃

kroggen:
It may fail with 70B. `unsigned long` or just `long` would fix it

ozabluda:
size_t was defined since C99, which is what llama.c is written in. `gcc -std=c99 llama.c` compiles, ` gcc -std=c89 llama.c` doesn't, not even close. llama.c is very very far from the original perfect K&R C :-). 

size_t is the unsigned integer type of the result of sizeof. In the first approximation, size_t is almost always a typedef to unsigned long long.

The "surgical fix" should really be unsigned long long, or maybe even better uint64_t.

run.c
```diff
@@ -132,11 +132,11 @@ void checkpoint_init_weights(TransformerWeights *w, Config* p, float* f, int sha
     w->rms_ffn_weight = ptr;
     ptr += p->n_layers * p->dim;
     w->w1 = ptr;
-    ptr += p->n_layers * p->dim * p->hidden_dim;
+    ptr += (unsigned int)p->n_layers * p->dim * p->hidden_dim;
```
ozabluda, line 135: this and below should really be unsigned long long, or maybe uint64_t. Possibly long long or int64_t,

ozabluda:
I just checked (`gcc -m32`) and on 32 bit platforms `ptr+=uint64_t` compiles. So it should be OK, especially with an assert.

kroggen:
hmmm, but these huge models will not work on 32-bit systems if using normal pointers anyway... (pointers are 32-bit there)

xefoci7612:
> #153 looks like has a lot going on. I'm really not a fan of this size_t stuff. I like int and float and such. Super simple. What the hell is size_t, it looks ugly.

@karpathy 

It happens that unfortunately memory addresses don't fit in a beautiful int nor in float...memory addresses fit in size_t, this is the reason why they felt the urge to invent it.

..and if you want to do pointer arithmetic (as we have plenty in run.c) or you go with size_t or you go casting as if there is no tomorrow.

I see why you don't like to change the original loaded types in Config, I understand why you want to avoid it, but the alternative IMHO seems more fragile....code correctness will depend on not forgetting some cast somewhere, not only in the current code, but in future patches too.

Anyhow it's up to you., of course.

karpathy:
closing as this PR fixed the issue, ty all https://github.com/karpathy/llama2.c/pull/367

PR #114, opened 2023-07-26
Master

rrajp:
Avoid division by Zero in decay ratio


karpathy:
warmup_iters should always be less than decay_iters... I'd probably prefer an assert or something.

PR #116, opened 2023-07-26
Cmake build support

krrishnarraj:
Cmake is cross platform make. Makes it easy to maintain single build file for almost all platforms.

* CmakeLists.txt file added
* Readme changes
* .gitignore file
* Since cmake prefers a separate build folder, the generated binaries will be in this `build` folder. To support loading `tokenizer.bin` file from non-root directory, a 4th argument is added to accept this filename


karpathy:
good to have this for anyone who wants it, will be able to find via search ty

PR #120, opened 2023-07-26
Handle failed time_in_ms

ianwal:
`time_in_ms()` returns -1 if there's an error, but this is never checked for which can lead to an inaccurate tok/s print message.


karpathy:
under what conditions does it return -1?

PR #121, opened 2023-07-27
stop generation on EOS or BOS

kroggen:
With this change, the model will generate a single "story" (or another content, depending on the model) per run

It appears that the basic models were not trained with EOS token, so the code also stops generation when it finds a BOS token


karpathy:
I don't want this behavior because I want the code to be dataset agnostic, and not have the concept of a "story" or something like that. Also it's easier in timing experiments to just always run for a fixed number of tokens.

PR #122, opened 2023-07-27
Add find_arg function and make the seed argument optional.

aegkmq:
Added a simple function to accepts various arguments.
For example, `find_arg("--seed", argv, argc)` returns the index+1 of last occurrence of `--seed` in `argv`, or 0 if it doesn't exist. It is possible that the returned index equals argc.


aegkmq:
Oh, the current implementation doesn't consider positional arguments. I'll open a new PR after the fix.

PR #124, opened 2023-07-27
Fix Windows timer, mmap, and enable openmp

richinseattle:
This uses mman-win32, GetTickCount, and updates the OMP pragmas to be MSVC compatible with minimal lines changed in run.c favoring including the mman-win32.c/.h. The ugliest part is probably the loop iterators need to be declared before the OMP pragma. Added Makefile target for mingw32 cross compile and a .bat file for compiling with cl.exe from MSVC.

### DIFF
--- a/run.c
+++ b/run.c
@@ -17,6 +17,8 @@ $ ./run
 #ifndef _WIN32
 #include <unistd.h>
 #include <sys/mman.h>
+#else
+#include "win.h"
 #endif 
 
 // ----------------------------------------------------------------------------
--- a/win.c
+++ b/win.c
@@ -1,0 +1,178 @@
+#include "win.h"
+#include <errno.h>
+#include <io.h>
+
+
+#ifndef FILE_MAP_EXECUTE
+#define FILE_MAP_EXECUTE    0x0020
+#endif /* FILE_MAP_EXECUTE */
+
+static int __map_mman_error(const DWORD err, const int deferr)
+{
+    if (err == 0)
+        return 0;
+    //TODO: implement
+    return err;
+}
+
+static DWORD __map_mmap_prot_page(const int prot)
+{
+    DWORD protect = 0;
+    
+    if (prot == PROT_NONE)
+        return protect;
+        
+    if ((prot & PROT_EXEC) != 0)
+    {
+        protect = ((prot & PROT_WRITE) != 0) ? 
+                    PAGE_EXECUTE_READWRITE : PAGE_EXECUTE_READ;
+    }
+    else
+    {
+        protect = ((prot & PROT_WRITE) != 0) ?
+                    PAGE_READWRITE : PAGE_READONLY;
+    }
+    
+    return protect;
+}
+
+static DWORD __map_mmap_prot_file(const int prot)
+{
+    DWORD desiredAccess = 0;
+    
+    if (prot == PROT_NONE)
+        return desiredAccess;
+        
+    if ((prot & PROT_READ) != 0)
+        desiredAccess |= FILE_MAP_READ;
+    if ((prot & PROT_WRITE) != 0)
+        desiredAccess |= FILE_MAP_WRITE;
+    if ((prot & PROT_EXEC) != 0)
+        desiredAccess |= FILE_MAP_EXECUTE;
+    
+    return desiredAccess;
+}
+
+void* mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off)
+{
+    HANDLE fm, h;
+    
+    void * map = MAP_FAILED;
+    
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable: 4293)
+#endif
+
+    const DWORD dwFileOffsetLow = (sizeof(off_t) <= sizeof(DWORD)) ? 
+                    (DWORD)off : (DWORD)(off & 0xFFFFFFFFL);
+    const DWORD dwFileOffsetHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
+                    (DWORD)0 : (DWORD)((off >> 32) & 0xFFFFFFFFL);
+    const DWORD protect = __map_mmap_prot_page(prot);
+    const DWORD desiredAccess = __map_mmap_prot_file(prot);
+
+    const off_t maxSize = off + (off_t)len;
+
+    const DWORD dwMaxSizeLow = (sizeof(off_t) <= sizeof(DWORD)) ? 
+                    (DWORD)maxSize : (DWORD)(maxSize & 0xFFFFFFFFL);
+    const DWORD dwMaxSizeHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
+                    (DWORD)0 : (DWORD)((maxSize >> 32) & 0xFFFFFFFFL);
+
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif
+
+    errno = 0;
+    
+    if (len == 0 
+        /* Unsupported flag combinations */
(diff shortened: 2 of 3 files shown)


karpathy:
@richinseattle , this PR now conflicts with @tairov PR here https://github.com/karpathy/llama2.c/pull/96 . I'm not an expert on this space, can you please describe the differences and why one might be preferred to the other? The one from @tairov looks a lot simpler.

Makefile
```diff
@@ -32,6 +32,10 @@ runfast: run.c
 runomp: run.c
 	$(CC) -Ofast -fopenmp -march=native run.c  -lm  -o run
 
+.PHONT: win64
```
karpathy, line 35: typo?

richinseattle:
#96 reverts my original timer patch and adds mmap code from a GPL project. I mine uses the simplest 1-line timer change, adds the most commonly used mmap replacement for win32 which is MIT licensed, and I enabled openmp.

Makefile
richinseattle: oops, fixed.

build_msvc.bat
```diff
@@ -0,0 +1 @@
+cl.exe /Ox /openmp /I. run.c mman-win32.c
```
karpathy, line 1: Sorry noob question but how is this file related to Makefile? What is this file for / is it needed?

run.c
```diff
@@ -190,8 +196,9 @@ void softmax(float* x, int size) {
 
 void matmul(float* xout, float* x, float* w, int n, int d) {
     // W (d,n) @ x (n,) -> xout (d,)
-    #pragma omp parallel for
-    for (int i = 0; i < d; i++) {
+    int i;
+    #pragma omp parallel for private(i)
```
karpathy, line 200: It doesn't work without the "for private(i)" part in Windows?

run.c
```diff
@@ -361,13 +369,17 @@ int argmax(float* v, int n) {
 
 long time_in_ms() {
     struct timespec time;
+#ifndef _WIN32
```
karpathy, line 372: should this be moved up one row? as `time` is not used in the windows version?

richinseattle: Yea, originally I was using my previous timer definition on windows which used timespec struct, but changed to GetTickCount since that is standard pattern on windows, can move the ifdef

run.c
richinseattle: correct, the counter variable has to be explicitly specified as private to each thread and declared before the pragma unfortunately.

build_msvc.bat
richinseattle: The Makefile is intended for cross compiling from linux, it will output a .exe. This .bat file would be what you run on windows if you have visual studio installed to generate an .exe from command line compiler cl.exe

run.c
richinseattle: I can use the approach from #96 to overload the clock_gettime definition in a win.h if you want me to clean up the timer related ifdef

karpathy: Yeah, we could see what that looks like. I don't really understand how the overloading happens though. By including the file up top, the function down below gets overwritten?

Makefile
```diff
@@ -32,6 +32,10 @@ runfast: run.c
 runomp: run.c
 	$(CC) -Ofast -fopenmp -march=native run.c  -lm  -o run
 
+.PHONY: win64
+win64: 
+	x86_64-w64-mingw32-gcc-win32 -Ofast -D_WIN32 -o run.exe -I. run.c mman-win32.c
```
karpathy, line 37: is it common to use underscores like this? why not just -DWIN32

richinseattle: _WIN32 is the compiler define, WIN32 is an application level define. I'm not sure if it is required with mingw tbh, it is defined automatically on MSVC.
see: https://stackoverflow.com/questions/662084/whats-the-difference-between-the-win32-and-win32-defines-in-c

tairov:
Thanks @richinseattle for this PR. As it was suggested in the issues, this is the variation that uses [mman-win32](https://github.com/alitrack/mman-win32) lib. 
The little downside is that it requires 2 additions.
I would keep the `mmap-windows` approach for simplicity, if "public domain" license is not a blocker of course. 
As a side note, I would recommend to split PRs for compilation errors fix & openmp support for windows platform.
It's nice to see contributions from Windows/C experienced people 😃

Makefile
karpathy: I see thank you.

run.c
richinseattle: I refactored and tested but it turns out that mingw doesn't have a timespec_get implementation so that would disallow cross-compilation. The best windows time_in_ms solution for time deltas is the ifdef I proposed here with GetTickCount()

karpathy: I'm not sure I understand. Isn't the timespec_get only used in the non-Windows version? So it should be moved inside the ifndef? It's spurious in Windows because we just call GetTickCount

richinseattle: On linux the function is clock_gettime(). My original patch proposed using timepsec_get because it exists on linux and MSVC and matches the same function prototype as clock_gettime(), and this was before I tried the cross-compile option which I think is convenient.

In theory the implementation could be hidden in a header that is only included on windows to implement clock_gettime() by using timespec_get() underneath. 

But that breaks mingw compiles, so we should stick with GetTickCount(), which does not return seconds+nano, so it needs to be placed in ifdefs in the time_in_ms() function since it returns ms resolution timing.

richinseattle:
@tairov I wasn't sure what happens when pd code is included in GPL, but it looks like it is safe to cherry pick.
https://www.gnu.org/licenses/gpl-faq.en.html#CombinePublicDomainWithGPL

If @karpathy prefers your malloc that's fine, it only implements mmap and munmap, vs the complete implementation in the mman-win32. We could flatten all windows support code beyond a couple ifdefs into a single .c and .h regardless since it's MIT.

If we go that route, I'll pull your mmap changes and then resubmit with the openmp support, however the timer solution still needs to be GetTickCount() as I mentioned above because timespec_get() breaks on mingw compiles.

tairov:
@richinseattle I totally agree and I got the point regarding `mingw & GetTickCount`. Isn't it worthwhile to create a separate PR for compatibility with `mingw` & `openmp`? It would be great to keep PRs atomic to the certain degree. As it should simplify reviews/reverts, etc.

tairov:
@richinseattle also a qq: do you think `llama2.c` need "complete implementation in the mman-win32"? What are the downsides to have implementations for mmap and munmap only?

richinseattle:
The complete API set is 
```
void*   mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off);
int     munmap(void *addr, size_t len);
int     mprotect(void *addr, size_t len, int prot);
int     msync(void *addr, size_t len, int flags);
int     mlock(const void *addr, size_t len);
int     munlock(const void *addr, size_t len);
```
I'm not strongly arguing you need the extra functions, just future proofing since it's not much extra code and standard practice for a mmap impl to have those APIs. There might be advantages to locking memory later (pretty sure they did that in llama.cpp).

richinseattle:
Regardless of which mmap, I would propose we just add a win.c/.h or win-support.c/.h (whatever name you prefer) and stuff anything else that comes up in the future in there in addition to the memory routines.

richinseattle:
Going to split this into 3 PRs

richinseattle:
Okay, openmp and timer support have merged. I have sent #130 if you want to use the mman-win32 impl, or #96 can be revised. Only need to add the 2 lines to run.c and decide which mmap. #96 is a single .h file, mine splits the .h and .c anticipating future needs. I will send a final PR to update build scripts after this is sorted

PR #125, opened 2023-07-27
Implemented sentencepiece BPE encoding and prompting

atamurad:
I've implemented sentencepiece encode function in C. Implementation required changes to tokenizer.py/tokenizer.bin to export piece scores. sentencepiece uses priority queues, but I used naive implementation to keep the new lines of code to minimal. 

I also added prompting. Interface and token injection pseudocode inspired by https://github.com/karpathy/llama2.c/pull/58 (thanks to @myan-o @leloykun )


run.c
```diff
@@ -371,15 +377,79 @@ long time_in_ms() {
...
+            return i;
+        }
+    }
+    return -1; // -1 if not found in vocabulary
+}
+
+int bpe_encode(char *text, Piece *vocab, int vocab_size, int *tokens, int *n_tokens)
+{
+    // encode text using vocabulary
+    char merged_str[128]; // max string size in vocabulary * 2
```
karpathy, line 394: i'm seeing 

>>> max(len(t) for t in tokens)
16

btw

karpathy: also shouldn't it be max_vocab_element_length*2  + 1 ? +1 for '\0'. So for us 16*2+1 = 33 maybe

run.c
```diff
@@ -371,15 +377,79 @@ long time_in_ms() {
...
+        if(id==-1) {
+            return 0;
+        }
+        tokens[*n_tokens] = id;
+        (*n_tokens)++;
+    }
+
+    // merge best consecutive pairs on each iteration
+    while(1) {
+        float best_score = -1e10;
```
karpathy, line 410: did you see that some of the scores in sentencepieve for some vocab elements is -1e9? What's up with that :S

run.c
atamurad: yes, I also confirmed that `max_sentencepiece_length: 16` in model train parameters. I will change it to 33.

karpathy: Even better, let's modify the .py script to export the max vocab element length? So we're not hardcoding?

run.c
atamurad: well, I've seen negative scores and -1e10 was just a guess. But looks like lucky one?

```
In [22]: s = spm.SentencePieceProcessor(model_file='tokenizer.model')

In [23]: min([s.get_score(i) for i in range(32000)])
Out[23]: -1000000000.0
```

karpathy: I guess I don't really understand what the -1e10 tokens are? Why are some tokens at random score -1e10

atamurad: No idea about scoring mechanism but these are tokens with -1e9 score in the model:
```
In [29]: [s.id_to_piece(i) for i in range(32000) if s.get_score(i)==-1e9]
Out[29]: 
['▁▁',
 '▁▁▁▁',
 '▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁',
 '▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁',
 '▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁',
 '▁']
```

run.c
atamurad: Pushed a commit with 33 for now. 
I'm also thinking if I can add a field to config struct (as vocab_size is already there) and populate it when loading the vocabulary, without exporting / adding a header to tokenizer.bin?

karpathy:
I ended up making many small changes, so I just submitted it as its own commit here https://github.com/karpathy/llama2.c/commit/b4bb47bb7baf0a5fb98a131d80b4e1a84ad72597 , building your work in this PR. Will close this one. Thank you!

atamurad:
> I ended up making many small changes, so I just submitted it as its own commit here [b4bb47b](https://github.com/karpathy/llama2.c/commit/b4bb47bb7baf0a5fb98a131d80b4e1a84ad72597) , building your work in this PR. Will close this one. Thank you!

Awesome, thank you for the commit!

PR #99, opened 2023-07-26
Improve locality

aegkmq:
Changing the order of two loops provides ~10% speed up. (`238 tok/s -> 260 tok/s`)

PR #133, opened 2023-07-27
Added Go ports

nikolaydubina:
Hello,

I have made few days ago Go port of this model. It is different from other Go port. It follows typical Go standard and notations better while preserving style of llama2.c

> I also think I did Go port first. haha !
> UPD: I see other versions of it popup: https://github.com/haormj/llama2.go so it is debatable who is first, if that matters anyways.

Anyways,

Would be great to include it here!

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ If your candidate PRs have elements of these it doesn't mean they won't get merg
 
 - [llama2.rs](https://github.com/gaxler/llama2.rs) by @gaxler: a Rust port of this project
 - [go-llama2](https://github.com/tmc/go-llama2) by @tmc: a Go port of this project
+- [llama2.go](https://github.com/nikolaydubina/llama2.go) by @nikolaydubina: a Go port of this project
 
 ## unsorted todos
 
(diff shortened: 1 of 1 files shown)


saracen:
I also created one: https://github.com/saracen/llama2.go

Separate from the others I guess in that it's dependency free, implements mmap, concurrency and I'm trying to eliminate bound checks where possible.

@karpathy Happy to add unless it feels like we're overcrowding the README with Go ports :smile:. Let me know.

karpathy:
gosh, haha!

tmc:
Gophers galore! Sounds like we need to get collaborating on one!

I believe mine is the most faithful line by line port and definitely want to increase performance while tracking fidelity to the c implementation.

kroggen:
The "bad locality" problem was fixed, but your ports all still have it

You can check [here](https://github.com/karpathy/llama2.c/pull/99/files) how it was made

nikolaydubina:
thanks. will look into it !

> your ports

haha. not really my ports ! I maintain only one

and original port (https://github.com/tmc/go-llama2/blob/master/go/run.go#L352) still has it too 🤷🏻‍♂️

---

speaking of solution, I see `memset`, Go does not have it. so need to see if we can adjust it for Go.

PR #135, opened 2023-07-27
README.md emoji upgrades

anupam-halder-india:
I add emojies to the Readme.md to make it look even good, pls don't ask way isa'nt there is 3 that 3rd commit, I just, just forgot the counting, OK?


karpathy:
GPT bots huh? it begins.

PR #138, opened 2023-07-27
Replace the rand() with a portable rng

### LINKED ISSUE #123: improved rand
`float r = (float)rand() / (float)RAND_MAX;` is a poor way to get random floats. Not only is `rand()` generally poor at generating entropy but the floats generated are biased and have more limited precision in many implementations.

The following code should produce better results
```c
// 32-bit pcg from https://en.wikipedia.org/wiki/Permuted_congruential_generator
#include <stdint.h>
static uint64_t       state      = 0x4d595df4d0f33173;
static uint64_t const multiplier = 6364136223846793005ull;
static uint64_t const increment  = 1442695040888963407ull;

static uint32_t rotr32(uint32_t x, unsigned r){
	return x >> r | x << (-r & 31);
}

uint32_t pcg32(void){
	uint64_t x = state;
	unsigned count = (unsigned)(x >> 59);		// 59 = 64 - 5

	state = x * multiplier + increment;
	x ^= x >> 18;					// 18 = (64 - 27)/2
	return rotr32((uint32_t)(x >> 27), count);	// 27 = 32 - 5
}

void pcg32_init(uint64_t seed){
	state = seed + increment;
	(void)pcg32();
}

// Produces floats in the  [0,1) range with uniform density
float rand_float(void){
    const uint32_t mask = 0x7fffff;
    const uint32_t sExp = 0x3f800000;
    union{
        uint32_t as_int;
        float as_float;
    }rnd = {
        .as_int = pcg32()
    };
    rnd.as_int &= mask;
    rnd.as_int |= sExp;
    return rnd.as_float - 1;
}
```

aegkmq:
close #123 , close #21


xefoci7612:
Sorry, I don't understand this line:

`return (random_u32() >> 9) * (1.0f / 8388608.0f);`

The conventional/idiomatic way to get a random float from a random integer seems to be:

`float x = (float)rand()/(float)RAND_MAX_64`

where in our case (64 bit):  `RAND_MAX_64 = 0x3FFFFFFF`


Maybe we can just use the native algorithm (without the last artificial right shift), and keep the conversion to float explicit as is now. Something like:

```C
#define RAND_MAX_64  0x3FFFFFFF

unsigned long long rng_seed;
unsigned int random_u64() {
    rng_seed ^= rng_seed >> 12;
    rng_seed ^= rng_seed << 25;
    rng_seed ^= rng_seed >> 27;
    return rng_seed * 0x2545F4914F6CDD1Dull;
}
.....

    // sample index from probabilities, they must sum to 1
    float r = (float)rand() / (float)RAND_MAX_64;
```

aegkmq:
[This paper](https://hal.science/hal-03282794/file/rand-in-range.pdf) says that to obtain a float in [0, 1), we should generate a random integer in the interval [0, 2^p - 1] and divide it by 2^p. (p=53 for double, p=24 for float)
So I'll modify p and keep the current method.

karpathy:
tbh it's a little bit scary to me that we're just winging it in a random PR. Makes me uncomfortable about correctness. Are there any verifiable references to some reputable repos that use this code nearly verbatim?

aegkmq:
I copied that conversion code from a previous version of [numpy](https://github.com/numpy/numpy/blob/859af34f7dd8524aad4d86a495944e20542815c7/numpy/random/src/distributions/distributions.c#L20). (They corrected the divisor for the [same reason](https://github.com/numpy/numpy/pull/20314))

m1lkweed:
> The conventional/idiomatic way to get a random float from a random integer seems to be:
> 
> `float x = (float)rand()/(float)RAND_MAX_64`
As #123 mentions, the idiomatic way isn't really good for generating random floats. I wrote [this](https://godbolt.org/z/jsxcx61fY) up in godbolt as an example. Numbers close to 1 are rounded to 1 and numbers near 0 drop down near 2e-10 with the idiomatic code. `rand_float()` produces more sensible values.

m1lkweed:
And [here](https://godbolt.org/z/4haqesodx) is an example with the code from this issue (`return (a >> 8) / 16777216.0f;`)

aegkmq:
It's just the difference between truncating methods. For example, `rand_float` returns the duplicate and non-increasing values in [this](https://godbolt.org/z/7Mefo14eE). And `(a & 0xffffff) / 16777216.0f` produces more sensible results than `rand_float` in [your test](https://godbolt.org/z/qxTT57fTs).

karpathy:
Thank you all, I'll take it.

cshenton:
Just FYI, xoshiro/xoroshiro PRNGs (such as the one I PR'd earlier in the week) are both faster and statistically more sound than xorshift PRNGs. Details: https://prng.di.unimi.it/ and https://xoshiro.di.unimi.it/hwd.php

If for whatever reason you happen to prefer a full 64 bit PRNG, even if 40 of those bits will be thrown away such as they are here, I'd at least suggest using xoshiro**/xoshiro++ PRNGs, which are now the default PRNG in .NET 6, Lua, Zig, and GNU Fortran.

Though, in the interest of balance, Javascript does use a Xorshift PRNG.

PR #143, opened 2023-07-27
Remove second ifdef by porting clock_gettime to win.c

tairov:
As there were discussions yesterday regarding further simplification of `run.c` I've created a PR with similar approach as in my PR #96 to remove `#if defined _WIN32` in `time_in_ms`.
The solution is to move WIN32 specific code to `win.c, win.h` in order to keep `run.c` clean.

Ported & functionally compatible version of `clock_gettime` moved to `win.c` 
```C
int clock_gettime(int clk_id, struct timespec *tp) {
    DWORD ticks = GetTickCount();
    tp->tv_sec = ticks / 1000;
    tp->tv_nsec = (ticks % 1000) * 1000000;
    return 0;
}
```

CI: https://github.com/tairov/llama2.c/actions/runs/5681971036

### DIFF
--- a/run.c
+++ b/run.c
@@ -365,15 +365,9 @@ int argmax(float* v, int n) {
 // ----------------------------------------------------------------------------
 
 long time_in_ms() {
-#if defined _WIN32
-    // windows specific way to get time
-    return GetTickCount();
-#else
-    // linux specific way to get time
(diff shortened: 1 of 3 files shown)


karpathy:
I like this trick, thank you!

PR #149, opened 2023-07-28
fix: Use correct compiler for Win64 GCC in Makefile

murilocurti:
**Fix: Use correct compiler for Win64 GCC in Makefile**

**Description:**
This PR fixes the Makefile to use the appropriate compiler for Win64 GCC. The original line used 'x86_64-w64-mingw32-gcc-win32', which is incorrect when using MSYS2. The correct executable name for MSYS2 is 'x86_64-w64-mingw32-gcc', which is now updated in the Makefile.

**Changes Made:**
- Modified the Makefile to use 'x86_64-w64-mingw32-gcc' for building on Win64 GCC.

**Testing:**
I tested the changes locally on my system to ensure that the Makefile builds the 'run.exe' executable correctly using the correct compiler.

Thanks for considering this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ runomp: run.c
 
 .PHONY: win64
 win64: 
-	x86_64-w64-mingw32-gcc-win32 -Ofast -D_WIN32 -o run.exe -I. run.c win.c
+	x86_64-w64-mingw32-gcc -Ofast -D_WIN32 -o run.exe -I. run.c win.c
 
 # compiles with gnu99 standard flags for amazon linux, coreos, etc. compatibility
 .PHONY: rungnu


karpathy:
@richinseattle can you take a look?

kroggen:
Maybe setup one job on the GitHub Actions workflow to test this, using the https://github.com/marketplace/actions/setup-msys2

richinseattle:
This is fine to pull, ubuntu distributions will indirectly call to the path I hardcoded but can be updated via deb-alternatives style schemes 
```
(base) rjohnson@codex:~/llama2$ ll /usr/bin/x86_64-w64-mingw32-gcc
lrwxrwxrwx 1 root root 40 Jul 26 02:16 /usr/bin/x86_64-w64-mingw32-gcc -> /etc/alternatives/x86_64-w64-mingw32-gcc*
(base) rjohnson@codex:~/llama2$ ll /etc/alternatives/x86_64-w64-mingw32-gcc
lrwxrwxrwx 1 root root 37 Jul 26 02:16 /etc/alternatives/x86_64-w64-mingw32-gcc -> /usr/bin/x86_64-w64-mingw32-gcc-win32*
```

tairov:
@kroggen I can take this. Just didn't get which OS `msys2` based on? windows or other?

kroggen:
It is for Windows. I just think it does not support arm, as is in the workflow, so maybe a separate job may be needed. Just x86_64 may be sufficient

richinseattle:
mingw is a gcc compiler with windows header/library support. It can run on linux or windows. 

MSYS2 is a unix-like shell environment built around mingw on windows. MSYS2 is out of scope here. 

The anticipation in my mind was for cross-compiles in a github release eventually as an add-on convenience when adding MSVC support. Currently, the MSVC build will be most performant because it supports openmp. I tested -fopenmp on the mingw compiler but it crashes when executed in the Windows environment.

This PR should be merged and the ARM topic should be opened seperately (I don't know anything about that)

murilocurti:
@richinseattle What do you think about add support for mingw, MSYS2 and MSVC?

The goal eith this is to give options to Windows developers choose.

ARM is something to research more here too.

Thanks

tairov:
I can confirm the correct executable name is `x86_64-w64-mingw32-gcc`, pushed a PR that implements CI step to verify compilations on `msys2/mingw` #175 
That PR also fixes executable name in `Makefile`

PR #153, opened 2023-07-28
Convert Config to use size_t

xefoci7612:
Use size_t instead of int in Config, this should fix a pointer arithmetic overflow with LLama 7B

Also fixed a real bug while there:

`float* weights_ptr = data + sizeof(Config)/sizeof(float); // float -> int !!!
`

### DIFF
--- a/run.c
+++ b/run.c
@@ -24,13 +24,13 @@ $ ./run
 // Transformer and RunState structs, and related memory management
 
 typedef struct {
-    int dim; // transformer dimension
-    int hidden_dim; // for ffn layers
-    int n_layers; // number of layers
-    int n_heads; // number of query heads
-    int n_kv_heads; // number of key/value heads (can be < query heads because of multiquery)
-    int vocab_size; // vocabulary size, usually 256 (byte-level)
-    int seq_len; // max sequence length
+    size_t dim; // transformer dimension
+    size_t hidden_dim; // for ffn layers
+    size_t n_layers; // number of layers
+    size_t n_heads; // number of query heads
+    size_t n_kv_heads; // number of key/value heads (can be < query heads because of multiquery)
+    size_t vocab_size; // vocabulary size, usually 256 (byte-level)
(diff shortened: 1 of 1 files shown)


kroggen:
I don't understand why these changes would fix the bug with 13B

Where exactly does the overflow happens?

I think it is not on the config

xefoci7612:
> I don't understand why these changes would fix the bug with 13B
> 
> Where exactly does the overflow happens?
> 
> I think it is not on the config

Yes, is not in the Config.

But Config fields are multiplied together leading to overflow, please see https://github.com/karpathy/llama2.c/pull/111 for a detailed description.

kroggen:
Is the overflow happening only on the `checkpoint_init_weights` function?

If yes, then this can be fixed with less changes, by using casting. eg:

```c
ptr += (size_t)p->n_layers * (size_t)p->dim * (size_t)p->hidden_dim;
```

kroggen:
Check #154

raimondasl:
@xefoci7612 @karpathy : I like this fix and change.

I ambitiously tried to see if this runs with 70B model. It fails, but I don't think it's the issue with the fix. Something is causing incorrect size calcs for w->freq_cis_real at least. I was wondering if that's because of the n_kv_heads and n_heads difference for 70B model. But there might be other differences that are causing export or import issues.

kroggen:
@raimondasl 

You can compile it with address sanitizer and undefined behavior sanitizer to check what is going on

More details on #164

karpathy:
@raimondasl yes exactly the 70B model uses multiquery, which run.c does not yet support.

xefoci7612:
I've analyzed memory usage. By far the biggest user is KV cache. I have verified it can overflow a 32 bit integer (>2GB) in LLama 7B.

All the other allocations are much smaller and there is no risk of overflowing even with a big LLama 70B.

In my repo I have verified there are very few places where you really need casting to size_t, so current patch is overkill and I close it.

PR #154, opened 2023-07-28
fix overflow with 13B model

kroggen:
Please someone test this

### DIFF
--- a/run.c
+++ b/run.c
@@ -118,32 +118,32 @@ void free_run_state(RunState* s) {
 void checkpoint_init_weights(TransformerWeights *w, Config* p, float* f, int shared_weights) {
     float* ptr = f;
     w->token_embedding_table = ptr;
-    ptr += p->vocab_size * p->dim;
+    ptr += (size_t) p->vocab_size * p->dim;
     w->rms_att_weight = ptr;
-    ptr += p->n_layers * p->dim;
+    ptr += (size_t) p->n_layers * p->dim;
     w->wq = ptr;
-    ptr += p->n_layers * p->dim * p->dim;
+    ptr += (size_t) p->n_layers * p->dim * p->dim;
     w->wk = ptr;
-    ptr += p->n_layers * p->dim * p->dim;
+    ptr += (size_t) p->n_layers * p->dim * p->dim;
     w->wv = ptr;
-    ptr += p->n_layers * p->dim * p->dim;
+    ptr += (size_t) p->n_layers * p->dim * p->dim;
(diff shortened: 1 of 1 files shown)


skeetmtp:
As stated in #164 , No more issue with this patch
There was an integer overflow, fixed by this change:
```
run.c:132:33: runtime error: signed integer overflow: 204800 * 13824 cannot be represented in type 'int'
```

run.c
```diff
@@ -222,7 +222,7 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
     float* freq_cis_imag_row = w->freq_cis_imag + pos * head_size / 2;
 
     // forward all the layers
-    for(int l = 0; l < p->n_layers; l++) {
+    for(size_t l = 0; l < p->n_layers; l++) {
```
ozabluda, line 225: Why is this necessary? p->n_layers is a small int, this introduces comparison of signed and unsigned, and for compiler optimizations it's better for loop counters to be signed.
```suggestion
    for(int l = 0; l < p->n_layers; l++) {
```

kroggen: It is the smallest change that fix the problem of overflow on places like `l*dim*dim`

ozabluda: Right. Then I think it's better to replace

`int dim = p->dim;`

with 

`size_t dim = p->dim;`

instead

kroggen: No, it must be the first element, so the next ones will be automatically casted to size_t on multiplication

ozabluda: I don't think so

> Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.

https://stackoverflow.com/questions/44055758/does-the-order-of-multiplication-variables-of-different-data-type-cause-differen

ozabluda: Just checked on my compiler (gcc) and it behaves exactly as I thought:

```
#include <limits.h>
#include <stddef.h>
#include <stdio.h>

int main()
{
  int x = INT_MAX;
  int z = INT_MAX;
  size_t y = 4;

  printf("%lu\n", x*y);
  printf("%lu\n", y*x);
  printf("%d\n", x*z);
}
```

```
$ gcc int.c -o int && ./int
8589934588
8589934588
1
```

You can also play with format, gcc warns about argument type mismatch, printing nicely the type of the product

karpathy:
closing as this PR fixed the issue, ty all https://github.com/karpathy/llama2.c/pull/367

PR #160, opened 2023-07-28
Add fp:fast for better performance on windows

GabrielJadderson:
without fp:fast     
achieved tok/s: 92.727273

with fp:fast    
achieved tok/s: 107.368421


karpathy:
@richinseattle are you able to reproduce?

richinseattle:
yep, tested and confirmed speed improvement. LGTM.

PR #161, opened 2023-07-28
Remove trailing spaces from python scripts

klei22:
Removing the trailing spaces from python files.

Just a small contribution at the moment, but does make it easier to work with with editors that highlight trailing spaces.

For these changes, relied on the `StripWhitespaces` command from the vim-plugin:

https://github.com/ntpeters/vim-better-whitespace


karpathy:
I think at some point would be good ty

PR #166, opened 2023-07-28
Running on GPU with OpenMP offloading

RahulSChand:
I have tried to add minimal changes in the code to run on gpu using openmp offloading. Currently only the matmul happens on GPU. Other operations are CPU.

There are few tmp variables introduced since offloading weights (i.e `w->k`) which point to virtual address & are only converted to actual address during runtime is not possible. 

I don't think this qualifies for a merge, but still open to suggestions.

### DIFF
--- a/run.c
+++ b/run.c
@@ -194,7 +194,10 @@ void softmax(float* x, int size) {
 void matmul(float* xout, float* x, float* w, int n, int d) {
     // W (d,n) @ x (n,) -> xout (d,)
     // by far the most amount of time is spent inside this little function
+    #pragma omp target data map(tofrom: w[0:d*n],x[0:n], xout[0:d])
+    {
     int i;
+    #pragma omp target device (0)
     #pragma omp parallel for private(i)
     for (i = 0; i < d; i++) {
         float val = 0.0f;
@@ -202,6 +205,7 @@ void matmul(float* xout, float* x, float* w, int n, int d) {
             val += w[i * n + j] * x[j];
         }
         xout[i] = val;
+    }
     }
 }
 
@@ -221,6 +225,11 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
     float* freq_cis_real_row = w->freq_cis_real + pos * head_size / 2;
     float* freq_cis_imag_row = w->freq_cis_imag + pos * head_size / 2;
 
+    //tmp files to read the current W being multiplied. Directly passing w->k .. doesn't work due to no support for virtual address offloading
+    float* tmp_w = (float*)malloc(dim*dim*sizeof(float));
+    float* tmp_w_hid = (float*)malloc(dim*hidden_dim*sizeof(float));
+    float* tmp_w_cls = (float*)malloc(p->vocab_size*dim*sizeof(float));
+
     // forward all the layers
     for(int l = 0; l < p->n_layers; l++) {
 
@@ -228,9 +237,12 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
         rmsnorm(s->xb, x, w->rms_att_weight + l*dim, dim);
 
         // qkv matmuls for this position
-        matmul(s->q, s->xb, w->wq + l*dim*dim, dim, dim);
-        matmul(s->k, s->xb, w->wk + l*dim*dim, dim, dim);
-        matmul(s->v, s->xb, w->wv + l*dim*dim, dim, dim);
+        memcpy(tmp_w, w->wq + l*dim*dim, dim*dim*sizeof(float));
+        matmul(s->q, s->xb, tmp_w, dim, dim);
+        memcpy(tmp_w, w->wk + l*dim*dim, dim*dim*sizeof(float));
+        matmul(s->k, s->xb, tmp_w, dim, dim);
+        memcpy(tmp_w, w->wv + l*dim*dim, dim*dim*sizeof(float));
+        matmul(s->v, s->xb, tmp_w, dim, dim);
 
         // apply RoPE rotation to the q and k vectors for each head
         for (int h = 0; h < p->n_heads; h++) {
@@ -300,7 +312,8 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
         }
 
         // final matmul to get the output of the attention
-        matmul(s->xb2, s->xb, w->wo + l*dim*dim, dim, dim);
+        memcpy(tmp_w, w->wo + l*dim*dim, dim*dim*sizeof(float));
+        matmul(s->xb2, s->xb, tmp_w, dim, dim);
 
         // residual connection back into x
         accum(x, s->xb2, dim);
@@ -310,8 +323,10 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
 
         // Now for FFN in PyTorch we have: self.w2(F.silu(self.w1(x)) * self.w3(x))
         // first calculate self.w1(x) and self.w3(x)
-        matmul(s->hb, s->xb, w->w1 + l*dim*hidden_dim, dim, hidden_dim);
-        matmul(s->hb2, s->xb, w->w3 + l*dim*hidden_dim, dim, hidden_dim);
+        memcpy(tmp_w_hid, w->w1 + l*dim*hidden_dim, hidden_dim*dim*sizeof(float));
+        matmul(s->hb, s->xb, tmp_w_hid, dim, hidden_dim);
+        memcpy(tmp_w_hid, w->w3 + l*dim*hidden_dim, hidden_dim*dim*sizeof(float));
+        matmul(s->hb2, s->xb, tmp_w_hid, dim, hidden_dim);
         
         // F.silu; silu(x)=x*σ(x),where σ(x) is the logistic sigmoid
         for (int i = 0; i < hidden_dim; i++) {
@@ -324,7 +339,8 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
         }
 
         // final matmul to get the output of the ffn
-        matmul(s->xb, s->hb, w->w2 + l*dim*hidden_dim, hidden_dim, dim);
+        memcpy(tmp_w_hid, w->w2 + l*dim*hidden_dim, hidden_dim*dim*sizeof(float));
+        matmul(s->xb, s->hb, tmp_w_hid, hidden_dim, dim);
 
         // residual connection
         accum(x, s->xb, dim);
@@ -334,7 +350,8 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
     rmsnorm(x, x, w->rms_final_weight, dim);
 
     // classifier into logits
-    matmul(s->logits, x, w->wcls, p->dim, p->vocab_size);
+    memcpy(tmp_w_cls, w->wcls, dim*p->vocab_size*sizeof(float));
+    matmul(s->logits, x, tmp_w_cls, p->dim, p->vocab_size);
 }
 
 // ----------------------------------------------------------------------------


kroggen:
The goal of OpenMP is to make the port simple, but if does not work with the memory pointers, I don't know if it helps that much

Do you have some benchmarks?

run.c
```diff
@@ -194,7 +194,10 @@ void softmax(float* x, int size) {
 void matmul(float* xout, float* x, float* w, int n, int d) {
     // W (d,n) @ x (n,) -> xout (d,)
     // by far the most amount of time is spent inside this little function
+    #pragma omp target data map(tofrom: w[0:d*n],x[0:n], xout[0:d])
+    {
     int i;
+    #pragma omp target device (0)
```
kroggen, line 200: cannot these 2 pragmas be joined into 1?
the `int i` can be removed, and use `for (int i = 0; i < d; i++) {`
also removing the `private(i)`

kroggen, line 200: cannot these 2 pragmas be joined into 1?
the `int i` can be removed, and use `for (int i = 0; i < d; i++) {`
also removing the `private(i)` and the additional `{` and `}`

kroggen: maybe this:
```c
    #pragma omp target map(tofrom: w[0:d*n],x[0:n], xout[0:d])
    for (int i = 0; i < d; i++) {
        float val = 0.0f;
        for (int j = 0; j < n; j++) {
            val += w[i * n + j] * x[j];
        }
        xout[i] = val;
    }
```
without the `parallel for` as well (if not going to be merged, if kept as separate repo)

run.c
```diff
@@ -194,7 +194,10 @@ void softmax(float* x, int size) {
 void matmul(float* xout, float* x, float* w, int n, int d) {
     // W (d,n) @ x (n,) -> xout (d,)
     // by far the most amount of time is spent inside this little function
+    #pragma omp target data map(tofrom: w[0:d*n],x[0:n], xout[0:d])
```
Ea0011, line 197: I am not an expert when it comes to CUDA but right now this is pretty basic as just mapping the data to GPU and using target directive is not going to utilize the GPU to the full extent. It would be interesting to see a benchmark as it might even be slower than the CPU version. Instead, to make this more optimal, `#pragma omp distribute` and `#pragma omp teams` should be used. What these will do is first create team of threads on the blocks of the GPU with `#pragma omp teams`. Then, the loop iterations will be distributed to the threads of the GPU with `#pragma omp distribute`. An example might look like this. Here, two loops are collapsed into 1, distributed to team of threads on the GPU.

```C
#pragma omp target teams distribute \
          parallel for collapse(2) map(to: x[0:n], w[0:d*n]) map(from: xout[0:d])
for (int i = 0; i < d; i++) { 
    for (int j = 0; j < n; j++) { 
          xout[i] += w[i * n + j] * x[j]; 
     } 
}
```

A good reference/guide from NVIDIA [OpenMP CUDA](https://on-demand.gputechconf.com/gtc/2018/presentation/s8344-openmp-on-gpus-first-experiences-and-best-practices.pdf)

run.c
```diff
@@ -221,16 +225,24 @@ void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights*
     float* freq_cis_real_row = w->freq_cis_real + pos * head_size / 2;
     float* freq_cis_imag_row = w->freq_cis_imag + pos * head_size / 2;
 
+    //tmp files to read the current W being multiplied. Directly passing w->k .. doesn't work due to no support for virtual address offloading
```
Ea0011, line 228: A thing to note with data is that moving it to and from GPU is an overhead. And we do it here in a loop by calling matmul in each iteration. And then in matmul we copy weights to the GPU. It would be efficient to try and map the whole weights to the GPU here before the loop by using `#pragma omp target data map` and then call matmul without having to copy data to GPU on each iteration. One can probably create a buffer for all weights, map it to GPU, then call matmul but I am not sure on this one. Just a direction worth exploring.

A reference here: [OpenMP CUDA](https://on-demand.gputechconf.com/gtc/2018/presentation/s8344-openmp-on-gpus-first-experiences-and-best-practices.pdf)

kroggen: I completely agree. It only makes sense if all memory is mapped on GPU memory

run.c
muralisubbarao2: Not an expert. Just taking a look at the code. I am not sure if the compiler optimizes computations in the nested for loops, like, e.g.,
for (int i = 0; i < d; i++) { 
   int in = i*n;
    for (int j = 0; j < n; j++) { 
          xout[i] += w[in + j] * x[j]; 
     } 
}

And if readability could be traded to reduce some computation, pointer arithmetic could be used
to access array elements instead of i and j;

run.c
RahulSChand: Thanks for the suggestions, will take a look

RahulSChand: Made two changes

1. Offloading all weights to GPU at once. Removed all the `tmp` variables defined

```
#pragma omp target data map(to:weights.wq[0:size_qkvO],weights.wk[0:size_qkvO],weights.wv[0:size_qkvO],weights.wo[0:size_qkvO],weights.w1[0:size_ffn],weights.w3[0:size_ffn],weights.w2[0:size_ffn],weights.wcls[0:size_cls])
while (pos < steps) {

// forward the transformer to get logits for the next token
transformer(token, pos, &config, &state, &weights);
....
```


2. Using variant of matmul loop in [comment](https://github.com/karpathy/llama2.c/pull/166#discussion_r1278164971). New matmul is below. Run state variables are still copied to GPU. 
```python
    int I;
    #pragma omp target data map(to:x[0:n]) map(from:xout[0:d])
    #pragma omp target teams distribute
    for (i = 0; i < d; i++) {
        float val = 0.0f;
        #pragma parallel for reduction(+:val) 
        for (int j = 0; j < n; j++) {
            val += w[i * n + j] * x[j];
        }
        xout[i] = val;
    }
}
```

Profiling. tok/s went from 20 to 80 currently. 
Next is try to allocate run state to GPU also & make the updates in it consistent with the value mapped to GPU.

Ea0011: Wow this was quick again :). Thank you, this is really interesting to follow. This code at least deserves a fork IMHO. I am curious, what are your specs and which model did you profile?

run.c
```diff
@@ -544,6 +549,11 @@ int main(int argc, char *argv[]) {
     int token = 1;   // init with token 1 (=BOS), as done in Llama-2 sentencepiece tokenizer
     int pos = 0;     // position in the sequence
     printf("<s>\n"); // explicit print the initial BOS token for stylistic symmetry reasons
+    int size_qkvO = config.dim*config.dim*config.n_layers;
+    int size_ffn = config.hidden_dim*config.dim*config.n_layers;
+    int size_cls = config.dim*config.vocab_size;
+
+    #pragma omp target data map(to:weights.wq[0:size_qkvO],weights.wk[0:size_qkvO],weights.wv[0:size_qkvO],weights.wo[0:size_qkvO],weights.w1[0:size_ffn],weights.w3[0:size_ffn],weights.w2[0:size_ffn],weights.wcls[0:size_cls])
```
Ea0011, line 556: Nice! Didn't really think it would be a single directive like this.

Ea0011 (commented):
Overall. This looks promising and very interesting at least to me. Really curios to see where this goes. Thank you again.

alexeadem:
Hello,

I compiled with icx to use my intel gpu in my fedora 38 laptop. I do see my gpu being used but I didn't see any improvement in performance 

# COMPILE

#cd $/HOMEintel/oneapi
#. ./setvars.sh
#cd -
(set -x; icx  -fiopenmp -fopenmp-targets=spir64 -fcf-protection=none  -fno-stack-protector -no-pie -O3 $1 -lm -o run)

# RUN
./draft/run llama2_7b.bin

I'm getting better performance with with base code without gpu offloading running it like this
```
 OMP_NUM_THREADS=4 ./run llama2_7b.bin
```

run.c
```diff
@@ -544,6 +549,11 @@ int main(int argc, char *argv[]) {
     int token = 1;   // init with token 1 (=BOS), as done in Llama-2 sentencepiece tokenizer
     int pos = 0;     // position in the sequence
     printf("<s>\n"); // explicit print the initial BOS token for stylistic symmetry reasons
+    int size_qkvO = config.dim*config.dim*config.n_layers;
+    int size_ffn = config.hidden_dim*config.dim*config.n_layers;
```
kroggen, line 553: Will these overflow with 7B or 13B models?

Maybe use `size_t`:
```c
size_t size_ffn = (size_t) config.hidden_dim * config.dim * config.n_layers;
```

run.c
kroggen: Where does the `weights.` prefix come from?

kroggen:
I am getting this error:

```
~/llama2.c-omp-cuda$ gcc -fopenmp -foffload=nvptx-none -fcf-protection=none  -fno-stack-protector -no-pie -O3 run.c -lm
ptxas fatal   : Value 'sm_30' is not defined for option 'gpu-name'
nvptx-as: ptxas returned 255 exit status
mkoffload: fatal error: x86_64-linux-gnu-accel-nvptx-none-gcc-9 returned 1 exit status
compilation terminated.
```

Using this docker image: [nvidia/cuda:12.2.0-devel-ubuntu20.04](https://hub.docker.com/r/nvidia/cuda/)

kroggen:
OK, I was able to build using image `nvidia/cuda:11.8.0-devel-ubuntu22.04`

And adding the `-foffload=-misa=sm_35` to the command line

```
git clone https://github.com/RahulSChand/llama2.c-omp-cuda
cd llama2.c-omp-cuda
wget https://huggingface.co/karpathy/tinyllamas/resolve/main/stories110M.bin
sudo apt update
sudo apt install -y gcc-offload-nvptx
gcc -fopenmp -foffload=nvptx-none -fcf-protection=none -foffload=-misa=sm_35 -fno-stack-protector -no-pie -O3 run.c -lm -o llama2
./llama2 stories110M.bin
```

But it is soooo slow, that I had to Ctrl+C it

run.c
```diff
@@ -195,14 +195,17 @@ void matmul(float* xout, float* x, float* w, int n, int d) {
     // W (d,n) @ x (n,) -> xout (d,)
     // by far the most amount of time is spent inside this little function
     int i;
-    #pragma omp parallel for private(i)
+    #pragma omp target data map(to:x[0:n]) map(from:xout[0:d])
+    #pragma omp target teams distribute
     for (i = 0; i < d; i++) {
         float val = 0.0f;
+        #pragma parallel for reduction(+:val)
```
Ea0011, line 202: Noticing a bug here. You are missing an `omp` in the directive. It should be `#pragma omp parallel for` not just `#pragma parallel for`. Also just a side note, using `#pragma omp simd` for the 2nd loop yielded much better results for me.

Ea0011:
@alexeadem Not much familiar with intel GPUs. What is the GPU that your laptop has? Also, there is a little bug that this code has. It is missing a keyword in a directive. So, what I think is happening is GPU is not utilized in its full potential now.

Ea0011:
@kroggen The code has a little bug on the line 202 :]. It states `#pragma parallel for reduction(+:val)` which misses an `omp` between parallel and pragma. 2nd loop essentially was not being parallelized at all I think.

kroggen:
I commented that line with `//` and the result was the same

Have you tried it? Is it also slow with you? In my test it is way slower than just CPU

alexeadem:
> @alexeadem Not much familiar with intel GPUs. What is the GPU that your laptop has? Also, there is a little bug that this code has. It is missing a keyword in a directive. So, what I think is happening is GPU is not utilized in its full potential now.

@Ea0011 
intel-gpu-top: Intel Alderlake_p (Gen12) @ /dev/dri/card1

Ea0011:
@kroggen Yes it was really slow for me. About 14 tok/s on 110M model which is not great. I replaced that line with `#pragma omp simd` and it dramatically improved. Now it is faster with 60tok/s than my laptop 6 core CPU. I am running it on the free colab with `T4` GPU by the way which isn't the fastest thing. I really think the bottleneck is synchronization between GPU and CPU states. I am thinking of ways to do more compute on GPU between data transfer periods.

I think we should use something like nvprof to profile the bottleneck though. I'll try to set it up on the cloud but am mot sure if this will work.

Ea0011:
@alexeadem you could also try changing line 202 with `#pragma omp simd reduction(+:val)` and see if that speeds things up for you.

alexeadem:
> @alexeadem you could also try changing line 202 with `#pragma omp simd reduction(+:val)` and see if that speeds things up for you.

hello @Ea0011 tried that and I get a warning:

```
:: initializing oneAPI environment ...
   build.sh: BASH_VERSION = 5.2.15(1)-release
   args: Using "$@" for setvars.sh arguments: run.c
:: compiler -- latest
:: debugger -- latest
:: dev-utilities -- latest
:: tbb -- latest
:: oneAPI environment initialized ::

/home/alex/git/llama2.c/draft
+ icx -fiopenmp -fopenmp-targets=spir64 -fcf-protection=none -fno-stack-protector -no-pie -O3 run.c -lm -o run
warning: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]

```
And there is no performance improvement. Maybe Intel doesn't support that...

Ea0011:
@alexeadem Interesting. I am really not familiar with intel stuff. On CUDA it did help but still this is not optimal still. I have a question though. Can I offload the same way you did for intel hd 600 series GPU s for instance or is it only for new GPUs?

kroggen:
In my case it is like 1 token/second. Cannot even wait it to finish printing

alexeadem:
> intel hd 600 series GPU

@Ea0011 it should

A quick test in Linux:

```
glmark2
sudo intel_gpu_top
```
And see if you gpu kicks in.

If it does you should be able to to use `icx` 
with 

icx -fiopenmp -fopenmp-targets=spir64
icpx -fiopenmp -fopenmp-targets=spir64

to enable openmp

leuc:
Offloading the weights to vram works really well, but the run-state struct might need some refactoring to fit OpemMPs data mapping requirements?

Even when limiting the offloading to the q, k, v I could not get valid results.

I suspect two issues. 

The call to matmul() with **variying** dimensions doesn't map correctly to the state in vram, when the state is offloaded early? 

The OpenMP memory boundaries have to be clearly defined with target enter and target exit in the transformer function. Or all computations on the state variables need to be offloaded, not just matmul().

Any ideas on how to improve the state struct for easier offloading?

ziliangpeng:
Is there still a chance this can be merged?

PR #167, opened 2023-07-29
Speed up tinystories pretokenize command

mzcu:
Brings massive speedup on systems with larger number of cores - on my 5950X, run time is cut down from around 1 hour to 2 minutes

### DIFF
--- a/tinystories.py
+++ b/tinystories.py
@@ -8,7 +8,7 @@ import json
 import os
 import random
 from typing import List
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import ProcessPoolExecutor
 
 import numpy as np
 import requests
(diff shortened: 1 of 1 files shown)


richinseattle:
This is great, took my pretokenize time from an hour to 38sec.

karpathy:
Is that juat because max_workers is hardwired to be 8 right now? Did you try removing that kwargs to get the system defaults, but still using threads?

karpathy:
appreciated ty

Majdoddin:
@karpathy 

> Is that juat because max_workers is hardwired to be 8 right now? Did you try removing that kwargs to get the system defaults, but still using threads?

No, @mzcu replaced ThreadPoolExecutor with ProcessPoolExecutor. Because of GIL, threading is mostly ineffective in python. 
Interestingly, they [decided](https://discuss.python.org/t/a-steering-council-notice-about-pep-703-making-the-global-interpreter-lock-optional-in-cpython/30474) for [nogil](https://github.com/colesbury/nogil) just a few days ago, but from python 3.13.

PR #170, opened 2023-07-29
Parallelize pretokenizer in multiple processes to boost performance

kefeiyao:
The python GIL (global interpreter lock) is holding back the parallelism of pretokenizer under ThreadPoolExecutor. Switch to ProcessPoolExecutor to eliminate GIL effect and enhance parallelism hence boost performance Unfortunately, the "right" parallel level setting depends on many things e.g., the total number of physical/virtual cores and the overall loading of the system etc. And too high setting may lock up the system. Thereby, add a new user setting for 'parallelism' for the user to choose at discretion with default set to 1 for safety.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ python tinystories.py download
 python tinystories.py pretokenize
 ```
 
+You can also change parallel setting of the pretokenizer for better performance by
+```bash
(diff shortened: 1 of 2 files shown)


karpathy:
merged a related alternative just now.

PR #171, opened 2023-07-29
Speed up BPE encoding

xefoci7612:
This PR speeds up BPE encoding of almost 20 times simplifying the code.

Specifically the number of calls to strcmp() inside str_lookup() are around 17 times less with this patch on a bunch of prompts. I expect this difference to grow with the length of the prompt.

**But this encoding do not use vocab scores!** 

Specifically we greedily encode consecutive tokens and iterate repeatedly until we don't find any more merges.

The speedup comes from the fact that current implementation discards all the possible token pairs if not the best one, then it will have to find them again. Current patch instead do not waste the job already done,.

I have briefly  compared the tokenization quality on around 10 different prompts and the tokens are almost the same both in number and in content. Actually with my surprise,  in many cases this patch even encodes with less tokens than the original.

I have not the background to understand how much it is important this vocab score in BPE. I looked at original BPE paper, but failed to find a clear answer on this specific point.

@karpathy if you are interested I can properly test with tinystories and measure the differences in tokenization between the 2 implementations so to have a proper test. Instead if you already know this is not good, could you please be so kind to point me to some paper/experiment so I can understand why. Thank you.

### DIFF
--- a/run.c
+++ b/run.c
@@ -350,8 +350,8 @@ int str_lookup(char *str, char **vocab, int vocab_size) {
     return -1;
 }
 
-void bpe_encode(char *text, char **vocab, float *vocab_scores, int vocab_size, unsigned int max_token_length, int *tokens, int *n_tokens) {
-    
+void bpe_encode(char *text, char **vocab, int vocab_size, unsigned int max_token_length, int *tokens, int *n_tokens) {
+
     // a temporary buffer to merge two consecutive tokens
     char* str_buffer = malloc((max_token_length*2+1) * sizeof(char)); // *2 for concat, +1 for null terminator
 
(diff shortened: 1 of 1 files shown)


BeBornTo:
Can you give some examples.
How did you measure a performance?

karpathy:
As this is probably not a big bottleneck right now and bloats the code, i will not accept it, but ty

karpathy:
"I have briefly compared the tokenization quality on around 10 different prompts and the tokens are almost the same both in number and in content. Actually with my surprise, in many cases this patch even encodes with less tokens than the original.
"

this is very worrying btw. the encoding has to be a deterministic function. i think this implementation is basically buggy. i'd compare to SentencePiece decoder in Python as the source of truth, and I expect there is a bug here that makes the outcomes disagree.

PR #175, opened 2023-07-29
Add build step for win64 msys2/mingw

tairov:
As it was suggested in the PR #149 I've implemented additional CI step for mingw (x86_64) compilation check.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,3 +94,31 @@ jobs:
         id: build_msvc
         run: |
           .\build_msvc.bat
+
+  windows-latest-mingw:
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: msys2 {0}
+
+    strategy:
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+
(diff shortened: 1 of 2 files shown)


kroggen:
Thanks! The cool thing about CI is that we can see that it is working

tairov:
@kroggen , exactly.. One additional step further we need in CI is a small test 😃 
In some cases I was experiencing binaries crashes, though compilation was successful. 

@karpathy do you have any idea how can we embed little test model into CI process to verify binaries?

kroggen:
If using temperature `0` and a fixed prompt, then it must generate the same output deterministically

So it can work as a regression test

The small 15M model could be downloaded. I just don't know about GitHub Actions limits

kroggen:
Maybe using the `cache` action to store the model?

https://github.com/actions/cache#example-cache-workflow

Then the jobs can check if the file is there on the sub-folder, and download it only if not

karpathy:
One idea is to generate the weights with a known rng seed, and then sample in a known-good implementation, and then expect that sequence of tokens?

PR #176, opened 2023-07-30
Switch to fast floating point model and turn off floating point exceptions for MSVC build

PeterReid:
Changing these build flags increases the achieved tokens per second on my computer from around 18 to around 21.


PeterReid:
I see this is already proposed in https://github.com/karpathy/llama2.c/pull/160 , so I am closing this.

Issue #165, opened 2023-07-28
mmap failed for llama2-7b on windows


tarunmcom:
succefully exported to .bin file from llama2-7b using the **export_meta_llama_bin.py** but getting error while running the file.
showing message-- mmap failed

richinseattle:
I will look tomorrow if someone else doesn't get to it

tarunmcom:
Thanks for replying some more info:-
1. The issue is with the llama2 converted bin file, other files eg stories42M.bin are working fine on windows.
2. Issue is while running on windows as the **same** llama2's bin file (model.bin) is working in Ubuntu
3. I tried to debug in Visual Studio even the line:- `file_size = ftell(file); // get the file size, in bytes` is returning -1

EdFries:
I was able to fix this by replacing that line with file_size = _ftelli64(file);
Be sure you are compiling for x64 or you will have other problems.

richinseattle:
One issue is that the `long` type on win64 is still 32bit. Really the codebase should switch to using stdint as there is another patch to fix integer overflows in another area of the code. I will wait until the other size patch gets committed before making one that works for windows. 

This specific change can be made minimally invasive for cross platform with this in `win.h` and changing file_size to be ssize_t type.
```
#define ssize_t __int64
#define ftell _ftelli64
```

richinseattle:
Fixed in https://github.com/karpathy/llama2.c/pull/179

PR #179, opened 2023-07-30
use ssize_t/int64 and 64bit version of ftell on windows

richinseattle:
On Windows, there is a separate API for ftell on 64bit, in addition due to long type being 32bit on Windows, switch to ssize_t

### DIFF
--- a/run.c
+++ b/run.c
@@ -483,7 +483,7 @@ int main(int argc, char *argv[]) {
     TransformerWeights weights;
     int fd = 0;         // file descriptor for memory mapping
     float* data = NULL; // memory mapped data pointer
-    long file_size;     // size of the checkpoint file in bytes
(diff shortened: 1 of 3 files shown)


karpathy:
Sorry can you say a bit more? Does the current code not work/run properly?

richinseattle:
This was a fix for https://github.com/karpathy/llama2.c/issues/165 This patch will successfully load llama2-7b after converting it. 
On Windows the `long` type is 32bit and the `ftell()` api is only 32bit, limiting use to 2GB files.

On Windows, the distinct api for 64bit ftell is `ftelli64()` so that was a simple update. But then we need to use a type that is appropriate for 64bit sizes with possible -1 return value, so `ssize_t` makes sense. I updated the mmap code to use 64bit sizes which simplified some of the logic, but technically it will still work on 32bit as well my commit comment was a little misleading.

PR #183, opened 2023-07-30
Add -funroll-all-loops to compiler flags

ozabluda:
Unroll all loops, even if their number of iterations is uncertain when the loop is entered. Runs faster on my tests.

For the difference in resulting assembly, see
https://godbolt.org/z/onaEsaEfT

Inspired by
https://github.com/karpathy/llama2.c/pull/95

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,27 +23,27 @@ rundebug: run.c
 # In our specific application this is *probably* okay to use
 .PHONY: runfast
 runfast: run.c
-	$(CC) -Ofast -o run run.c -lm
+	$(CC) -Ofast -funroll-all-loops -o run run.c -lm
 
 # additionally compiles with OpenMP, allowing multithreaded runs
 # make sure to also enable multiple threads when running, e.g.:
 # OMP_NUM_THREADS=4 ./run out/model.bin
 .PHONY: runomp
 runomp: run.c
-	$(CC) -Ofast -fopenmp -march=native run.c  -lm  -o run
+	$(CC) -Ofast -funroll-all-loops -fopenmp -march=native run.c  -lm  -o run
 
 .PHONY: win64
 win64: 
-	x86_64-w64-mingw32-gcc-win32 -Ofast -D_WIN32 -o run.exe -I. run.c win.c
+	x86_64-w64-mingw32-gcc-win32 -Ofast -funroll-all-loops -D_WIN32 -o run.exe -I. run.c win.c
 
 # compiles with gnu99 standard flags for amazon linux, coreos, etc. compatibility
 .PHONY: rungnu
 rungnu:
-	$(CC) -Ofast -std=gnu11 -o run run.c -lm
+	$(CC) -Ofast -funroll-all-loops -std=gnu11 -o run run.c -lm
 
 .PHONY: runompgnu
 runompgnu:
-	$(CC) -Ofast -fopenmp -std=gnu11 run.c  -lm  -o run
+	$(CC) -Ofast -funroll-all-loops -fopenmp -std=gnu11 run.c  -lm  -o run
 
 .PHONY: clean
 clean:


kroggen:
Interesting. It does not change the code.

Please post the result of some benchmarks, without and with this flag, for each makefile section

It is important also to compare when using `OMP_NUM_THREADS=4 ...`

karpathy:
I'm not able to reproduce a speedup here on my Linux box. Do you have some timings to share? Also I can't seem to compile this with clang, only with gcc. A warning gets issued that this flag is unsupported.

ozabluda:
on my Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz with `./run stories110M.bin`

```
runfast:
tok/s: 33.337691 (without -funroll-all-loops)
tok/s: 33.860045 (with -funroll-all-loops) (+1.6%)

runopenmp:
tok/s: 48.067861 (without -funroll-all-loops)
tok/s: 48.925556 (with -funroll-all-loops) (+1.8%)
```

To see the possible slight performance improvement on your particular box, you have to run it a couple of times and take the _max_, as it corresponds to the actual performance sans noise (noise is always additive).

Slight performance improvement is in line with slight performance improvement reported here from manual hand-unrolling: 
https://github.com/karpathy/llama2.c/pull/95#issuecomment-1650970413

In some cases improvement from may be better. @krzysztof-jusiak reported huge performance improvement with manual hand-unrolling on his use case. This PR was meant is a just a probably much better alternative to the proposed hand-unrolling. 

@kroggen, what exactly do you mean "It does not change the code."? In godbolt link above it clearly does, even compared to hand-unrolling from the PR that inspired this PR, let alone original vanilla.

ozabluda:
I agree that on the platforms tested so far it's barely worth it, maybe if more performance improvement is reported by others it'll be worth it. 

This PR was inspired by https://github.com/karpathy/llama2.c/pull/95 where it was reported that with hand-unrolling

> Result (with float16):
> 
> before: 16tok/s
> after: 71tok/s

I am waiting for something like https://github.com/karpathy/llama2.c/pull/93 to merged to recheck and possibly reopen this PR.

twobob:
-funroll-all-loops isnt supported on clang anyway. AFAIK

clang -Ofast -fopenmp -funroll-all-loops -D_WIN32 -fprofile-instr-use=default.profdata -o run_clang.exe -I. run.c win.c
clang: warning: optimization flag '-funroll-all-loops' is not supported [-Wignored-optimization-argument]
clang: warning: optimization flag '-funroll-all-loops' is not supported [-Wignored-optimization-argument]

twobob:
to be fair. I run the 4x manual unroll for now anyway, but it would be nice for others.

ozabluda:
Try with 8. That's what gcc does (see godbolt above). How much speedup do you get? A couple percent, like everybody else?

twobob:
I went with 
win64: 
	x86_64-w64-mingw32-gcc -Ofast  -funroll-all-loops  -fopenmp -DCOMPILER=\"MINGW\"  -D_WIN32 -o runmingw.exe -I. run.c win.c
winclang: 
	clang -march=native -fno-math-errno -fopenmp -DCOMPILER=\"CLANG\" -D_CRT_SECURE_NO_WARNINGS   -Ofast -D_WIN32 -o run.exe -I. run.c win.c
wingcc: 
	$(CC) -march=native -fno-math-errno  -funroll-all-loops -fopenmp -DCOMPILER=\"GCC\"  -Ofast -D_WIN32 -o rungcc.exe -I. run.c win.c

achieved tok/s: 40.514851 for MINGW
achieved tok/s: 40.742363 for GCC
achieved tok/s: 43.186424 for CLANG

which is a marked improvement for mingw and gcc over the previously ubiquitously superior clang build; what might have been as much as a 15% deficits sometimes clawed back within almost noise levels. Windows. Junky machine. 
thanks for the tip. And for the godbolt

PR #187, opened 2023-07-31
Adding LoRA fine tuning

wlamond:
Adding an implementation of LoRA fine tuning, heavily inspired by [minLoRA](https://github.com/cccntu/minLoRA). I thought the use of pytorch parametrization was interesting and simple, and fits in nicely with the approach of this project. Let me know if you were thinking of explicitly implementing the modified forward pass rather than a factored/merged forward pass, or if you think this would be a better fit as a separate repo.

I added the tinyshakespeare dataset and default to fine tuning on that. I wanted to tune the tinystories models a small amount (~50-100 steps) to get Shakespearian tiny stories :) I had some mixed results, e.g.:

```Once upon a time, there was a boy named Tom and a girl named Lily. They lived in a big, expensive house, but they did not need each other. Tom and Lily were always together, even on the same summer.
Tom and Lily played every day, but they did not get along very well. They would fight, argue, and not be friends. They were angry at each other and their answers. Their hearts were hopeful but unfat.
Finally, after all their plays, they said to each other:
"I thought you would be mad, I don't really want to be your friend!"
At first, they felt sad, but as they began to talk more, neither of them had the same thought and the same word. In the end, he was in a new state yet he was still alive.```

Still mostly story-like, but certainly leaning more towards the drama of Shakespeare. I like the commentary on how being exposed to new and original thoughts can leave you in a new state of being. ;)

I also tuned this for ~1k steps with the 15M param model to get something that more closely resembles Shakespeare.

I only have access to a 1080ti and a v100 16GB, so I wasn't able to do more thorough testing/experimentation on the actual Llama2 checkpoints. Let me know if you'd like to see more testing before making a decision on what to do with this. 

Thanks for sharing this project! It's been fun to play with.

### CODE REVIEW

**Comment by Ea0011** [comment]
This looks elegant to me.

**Comment by karpathy** [comment]
I like where this is going, but this looks like multiple PRs in one, and a little bit of sus code. I'll inline comment

File: train.py [hunk 1]
```diff
@@ -170,10 +162,35 @@
...
+    out_dir = out_dir + "_lora_finetune"
+    os.makedirs(out_dir, exist_ok=True)
+    for p in model.parameters():
+        p.requires_grad = False
+    apply_lora(model, layer_types=[nn.Linear], dropout=dropout)
+    # Optionally tie the LoRA weights between the embedding layer and the language model output.
+    # I found this didn't help much for Shakespeare, but it might help for other tasks.
+    # tie_lora_weights(model.output, model.tok_embeddings)
+
+    # replace our Task dataset class with the Tiny Shakespeare one.
```
karpathy, line 179: not looking good here. hardcoding?

wlamond: Whoops, yeah that's janky. I'll make it configurable. 

Re: looking like multiple PRs, are you referring to the addition of the tinyshakespeare dataset + LoRA? I'd be happy to separate these into two PRs

vgoklani:
@wlamond If we're going to `LoRA`, then why not just go all the way and do `QLoRA` :)

It's a very simple change to your PR, just need to reference bnb.nn.Linear4bit for the 4-bit quantization.

wlamond:
> @wlamond If we're going to `LoRA`, then why not just go all the way and do `QLoRA` :)
> 
> It's a very simple change to your PR, just need to reference bnb.nn.Linear4bit for the 4-bit quantization.

@vgoklani Oooo, I do like that idea. I think it would be better as a separate PR though. I'm not sure how Andrej feels about adding other dependencies, so I'd rather get this project finished and then add QLoRA as another option if there's interest. Thanks for the idea and feedback!

train.py
wlamond: I removed all the tinyshakespeare code from this branch + PR and submitted it as a separate PR (#211). This one can merge on its own, so folks can try fine tuning the pretrained Llama2s on tinystories, or we can tune the tinystories models on Shakespeare if both get merged.

twobob:
> > @wlamond If we're going to `LoRA`, then why not just go all the way and do `QLoRA` :)
> > It's a very simple change to your PR, just need to reference bnb.nn.Linear4bit for the 4-bit quantization.
> 
> @vgoklani Oooo, I do like that idea. I think it would be better as a separate PR though. I'm not sure how Andrej feels about adding other dependencies, so I'd rather get this project finished and then add QLoRA as another option if there's interest. Thanks for the idea and feedback!

definitely interest. adds possibilities to potentially do more with less 🙇‍♂️

train.py
```diff
@@ -173,8 +180,19 @@
...
+    if init_from == 'resume':
+        iter_num = checkpoint["iter_num"]
+        best_val_loss = checkpoint["best_val_loss"]
+
+if init_from == "lora_finetune":
+    out_dir = out_dir + "_lora_finetune"
+    os.makedirs(out_dir, exist_ok=True)
+    for p in model.parameters():
+        p.requires_grad = False
+    apply_lora(model, layer_types=lora_layer_types, rank=lora_rank, dropout=lora_dropout, alpha=lora_alpha)
```
ecr23xx, line 192: Looks like you apply LoRA to every Linear layer? Could you provide the ability to register lora for target modules (e.g. wq and wk only as suggested in the original paper)? The code may look like:

```python
def apply_lora(model, ..., target_modules=['wq', 'wk']:
    for name, layer in model.named_modules():
        if name.split(".")[-1] not in target_modules:
            continue
        # register lora parameterization

```

train.py
```diff
@@ -332,5 +350,12 @@ def get_lr(it):
     if iter_num > max_iters:
         break
 
+if init_from == "lora_finetune":
+    print('merging lora')
+    merge_lora(raw_model)
```
ecr23xx, line 355: Maybe save lora parameters in a standalone file?

ecr23xx:
Another possible improvement: the original parameters doesn't need to be stored in the optimizer during lora finetuning.

train.py
wlamond: I love this idea! I have a local implementation that does that, but I'll update this to follow suit.

train.py
wlamond: Agreed, the checkpoints actually already have the lora parameters in them (the parameterization is computed whenever the weights are referenced, including during exports). Saving them off to the side could enable hot swapping loras for different tasks at some point if folks are interested in that feature.

ecr23xx:
> > @ecr23xx Another possible improvement: the original parameters doesn't need to be stored in the optimizer during lora finetuning.
> 
> The `configure_optimizers` method only passes parameters with `requires_grad == True` to the optimizer and it's called after we set up lora and freeze the original weights, so we should be all set here!

Oh I got it. Look forward to your updates 🚀

model.py
```diff
@@ -388,3 +389,55 @@ def serialize(t):
...
+        nn.init.kaiming_uniform_(self.lora_a, a=math.sqrt(5))
+
+        self.scaling = alpha / rank
+        self.dropout = nn.Dropout(dropout_p)
+
+    def forward(self, weight):
+        return weight + torch.matmul(self.lora_b, self.dropout(self.lora_a)) * self.scaling
+
+
+def apply_lora(model: nn.Module, layer_types=[nn.Linear], rank=8, dropout=0.0, alpha=1.0):
```
karpathy, line 424: I always get a little antsy seeing Lists in defaults

https://docs.python-guide.org/writing/gotchas/#what-you-wrote

model.py
```diff
@@ -388,3 +389,55 @@ def serialize(t):
...
+    def _apply_lora(module):
+        if type(module) in layer_types and hasattr(module, 'weight'):
+            fan_out, fan_in = module.weight.shape
+            parametrize.register_parametrization(module, 'weight', LoraLinear(fan_in, fan_out, rank, dropout, alpha))
```
karpathy, line 428: would this fail if LayerTypes is nn.Embedding? It shouldn't get replaced with LoraLinear right?

train.py
```diff
@@ -173,8 +180,19 @@
...
             state_dict[k[len(unwanted_prefix) :]] = state_dict.pop(k)
     model.load_state_dict(state_dict)
-    iter_num = checkpoint["iter_num"]
-    best_val_loss = checkpoint["best_val_loss"]
+    if init_from == 'resume':
+        iter_num = checkpoint["iter_num"]
+        best_val_loss = checkpoint["best_val_loss"]
+
+if init_from == "lora_finetune":
+    out_dir = out_dir + "_lora_finetune"
```
karpathy, line 188: os.path.join?

gohai:
@wlamond I'd love to do some experimentation with LoRA on various types of smaller models. Any chance this PR could be revived/updated?

PR #189, opened 2023-07-31
clang vectorisation info update

mrsteyk:
This PR adds some clarification as to why clang might produce worse results than gcc. I was not able to find proper documentation for this. Best I could find was [this Google Groups thread.](https://groups.google.com/g/llvm-dev/c/uoSFWMozu2M/m/04XQQ1lSAgAJ)


karpathy:
i appreciate the PR

PR #82, opened 2023-07-25
End on BOS token

xefoci7612:
End generation on next BOS token

PR #190, opened 2023-07-31
Terminate the chat if EOS token is encountered

bjmsong:


twobob:
Yeah I already do this. but people may want that to be an option rather than a baked-in business rule. dunno.

kroggen:
Duplicate of #82 and #121

PR #191, opened 2023-07-31
Handle mid-token issue

xefoci7612:
Fix the so called "mid-token" issue by stripping the last token from the prompt and then sampling the first processed token after the prompt among the ones that start with the last token's string.

We also got a new feature for free. We can now explicitly set last word boundary in the prompt, i.e. signal if the last word of the prompt should be considered a full word or just a partial word.

For instance, in this case:

```
./run model.bin 0 5 "On"
<s>
One day, a little
```

Last word of the prompt is expanded On -> One by the model. Instead if we want the sentence to specifically start with the word "On" we can now simply submit:

```
./run model.bin 0 5 "On "
<s>
On a hot summer day
```

### DIFF
--- a/run.c
+++ b/run.c
@@ -399,6 +399,16 @@ void bpe_encode(char *text, char **vocab, float *vocab_scores, int vocab_size, u
     free(str_buffer);
 }
 
+void token_candidates(int* excluded_tokens, int last_token, char **vocab, int vocab_size) {
+    char* last = vocab[last_token];
+    size_t len = strlen(last);
+    for (int i = 0; i < vocab_size; i++) {
+        if (strncmp(last, vocab[i], len) != 0) {
+            excluded_tokens[i] = 1;
+        }
+    }
+}
+
 // ----------------------------------------------------------------------------
 // utilities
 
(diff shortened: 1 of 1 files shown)


karpathy:
Tempting to merge this though it does bloat the complexity a bit with a fairly subtle issue sad. I think I will merge this though... 

Btw here is a related thing on the topic https://github.com/microsoft/guidance/blob/main/notebooks/token_healing.ipynb

This is called "token healing"

xefoci7612:
@karpathy thanks for the link

Now that Top-p sampling has been merged a straightforward approach (with minimal number of LOC added) would be

Top-p -> add Top-K option too (I did it in my repo, a very minimal change) -> add token healing (naturally fits in Top-K logic after setting all the other logits to a minimum)

karpathy:
@xefoci7612 do you think we should be able to just clamp the logits at the excluded locations to -inf? Instead of taking it into the sampling code.

xefoci7612:
@karpathy 

This is another attempt according to your suggestion. It is compatible with _any_ kind of sampling, argmax, top-p, etc because tokens clamping is done upstream at logits level.

#248

PR #195, opened 2023-07-31
Adjust `malloc` size for `prompt_tokens`

clebert:
This pull request addresses a potential memory optimization when calling the `bpe_encode` function.

The function currently reserves memory for the `prompt_tokens` array using a size derived from `config.seq_len`, independent of the actual size of the `prompt`. This could lead to a situation where the allocated memory size is either too large (leading to inefficient usage of memory) or even too small if the `prompt` has more characters than `config.seq_len` (potentially resulting in buffer overflow).

This PR proposes to change the `malloc` call from
- `(int*)malloc(config.seq_len * sizeof(int))` to 
- `(int*)malloc(strlen(prompt) * sizeof(int))`

This ensures that the allocated memory is directly proportional to the size of the `prompt`.

### DIFF
--- a/run.c
+++ b/run.c
@@ -534,7 +534,7 @@ int main(int argc, char *argv[]) {
     int *prompt_tokens = NULL;
     int num_prompt_tokens = 0;
     if (prompt != NULL) {
-        prompt_tokens = (int*)malloc(config.seq_len * sizeof(int));
+        prompt_tokens = (int*)malloc(strlen(prompt) * sizeof(int));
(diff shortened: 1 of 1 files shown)


mrsteyk:
2048*4 is 8kb, model takes more space than this temporary buffer. Try tokenising [tokenizer.py](https://raw.githubusercontent.com/karpathy/llama2.c/master/tokenizer.py) and comparing it's size. This is just useless...

karpathy:
thank you, i agree, esp the concern w.r.t. potential buffer overflow.

PR #205, opened 2023-08-01
add Top-p sampling

bjmsong:
I asked the same questions, the top-p can get better answer.
Model： llam2_7b_chat

before:
image

image

after:
image

image

### DIFF
--- a/run.c
+++ b/run.c
@@ -510,7 +510,7 @@ int main(int argc, char *argv[]) {
     TransformerWeights weights;
     int fd = 0;         // file descriptor for memory mapping
     float* data = NULL; // memory mapped data pointer
-    long file_size;     // size of the checkpoint file in bytes
+    ssize_t file_size;     // size of the checkpoint file in bytes
     {
         FILE *file = fopen(checkpoint, "rb");
         if (!file) { printf("Couldn't open file %s\n", checkpoint); return 1; }
--- a/run.c
+++ b/run.c
@@ -446,10 +446,10 @@ int sample_top_p(float* probs, int n, float topp) {
         accum += probs_sort[p].value;
     }
 
-    float r = random_f32();
+    float r = random_f32() * accum;
     float cdf = 0.0f;
     for (int i = 0; i < p + 1; i++) {
-        cdf += probs_sort[i].value/accum;
+        cdf += probs_sort[i].value;
         if (r < cdf) {
             return probs_sort[i].index;
         }
--- a/run.c
+++ b/run.c
@@ -421,16 +421,49 @@ float random_f32() { // random float32 in [0,1)
     return (random_u32() >> 8) / 16777216.0f;
 }
 
-int sample(float* probabilities, int n) {
-    // sample index from probabilities, they must sum to 1
-    float r = random_f32();
-    float cdf = 0.0f;
+typedef struct {
+    float value;
+    int index;
+} IndexedFloat;
+
+int compare_indexed_float(const void* a, const void* b) {
+    IndexedFloat* indexed_a = (IndexedFloat*)a;
+    IndexedFloat* indexed_b = (IndexedFloat*)b;
+    // sorting in descending order
+    return (indexed_b->value > indexed_a->value) ? 1 : ((indexed_b->value < indexed_a->value) ? -1 : 0);
+}
+
+int sample_top_p(float* probs, int n, float topp) {
+    IndexedFloat probs_sort[n];
     for (int i = 0; i < n; i++) {
-        cdf += probabilities[i];
-        if (r < cdf) {
-            return i;
+        probs_sort[i].value = probs[i];
+        probs_sort[i].index = i;
+    }
+
+    qsort(probs_sort, n, sizeof(IndexedFloat), compare_indexed_float);
+
+    // Calculate cumulative probabilities
+    float cum_probs[n];
+    cum_probs[0] = probs_sort[0].value;
+    int p = 1;
+    for (; cum_probs[p-1]<=topp && p < n; p++) {
+        cum_probs[p] = cum_probs[p - 1] + probs_sort[p].value;
+    }
+
+    for (int i = 0; i < p; i++) {
+        probs_sort[i].value /= cum_probs[p-1];
+    }
+
+    // Sample next token based on the modified probability distribution
+    float r = random_f32();
+    float cum_prob = 0.0;
+    for (int i = 0; i < p; i++) {
+        cum_prob += probs_sort[i].value;
+        if (cum_prob > r) {
+            return probs_sort[i].index;
         }
     }
+
     return n - 1; // in case of rounding errors
 }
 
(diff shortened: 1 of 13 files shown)


karpathy:
This does more than just top-p, e.g. introduces EOS truncation. The code diff is also too large, maybe there is something more minimal / simple? Is qsort sufficiently cross-platform?

twobob:
> This does more than just top-p, e.g. introduces EOS truncation. The code diff is also too large, maybe there is something more minimal / simple? Is qsort sufficiently cross-platform?

    Linux: qsort is widely supported on Linux, since the qsort function is part of the C Standard Library and is defined in the stdlib.h header.
    Windows: qsort is supported by the Microsoft C Runtime Library
    macOS: qsort is supported on macOS again with a declaration in the stdlib.h header file, and its behavior should be consistent with other platforms.
    
IndexedFloat and compare_indexed_float seem correct as the code uses standard C constructs and should be portable across Linux, Windows, and macOS without any platform-specific issues

The road to hell... but AFAIK it should be okay

run.c
```diff
@@ -421,16 +421,49 @@ float random_f32() { // random float32 in [0,1)
...
+    // Calculate cumulative probabilities
+    float cum_probs[n];
+    cum_probs[0] = probs_sort[0].value;
+    int p = 1;
+    for (; cum_probs[p-1]<=topp && p < n; p++) {
+        cum_probs[p] = cum_probs[p - 1] + probs_sort[p].value;
+    }
+
+    for (int i = 0; i < p; i++) {
+        probs_sort[i].value /= cum_probs[p-1];
```
tringwald, line 454: I think this rescaling can be avoided by scaling the sampled float `r` with `cum_probs[p-1]` instead.

run.c
```diff
@@ -421,16 +421,49 @@ float random_f32() { // random float32 in [0,1)
...
+    float r = random_f32();
+    float cum_prob = 0.0;
+    for (int i = 0; i < p; i++) {
+        cum_prob += probs_sort[i].value;
+        if (cum_prob > r) {
+            return probs_sort[i].index;
         }
     }
+
     return n - 1; // in case of rounding errors
```
tringwald, line 467: Not sure if it makes a difference, but shouldn't this be `p-1`? Because at this point, we already know that the sum up until the `p-th` element is larger than `topp` and therefore `p-1` should be that last valid index to return.

bjmsong:
Thanks to @twobob @tringwald 
update：
- code simplification
- remove EOS truncation

run.c
```diff
@@ -421,16 +421,49 @@ float random_f32() { // random float32 in [0,1)
...
-        if (r < cdf) {
-            return i;
+        probs_sort[i].value = probs[i];
+        probs_sort[i].index = i;
+    }
+
+    qsort(probs_sort, n, sizeof(IndexedFloat), compare_indexed_float);
+
+    // Calculate cumulative probabilities
+    float cum_probs[n];
```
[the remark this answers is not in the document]
bjmsong: thanks，I have updated this piece of code.

run.c
twobob: Suppose we have the following probability distribution and top-p threshold:

    probs = [0.1, 0.2, 0.3, 0.4]
    topp = 0.7

We want to sample an index from the top 70% of probabilities.
Working Through the Code

    Sort and Accumulate: The code will sort the probabilities and accumulate them until the accumulated value exceeds topp. In this case:
        probs_sort = [(0.4, 3), (0.3, 2), (0.2, 1), (0.1, 0)]
        The accumulation will stop at p = 3 since 0.4+0.3+0.2=0.9>0.70.4+0.3+0.2=0.9>0.7.

    Sampling Loop: The loop will consider probabilities from index 0 to p. In this example, it will consider indices 0, 1, and 2.

    Rounding Errors: If, due to rounding errors, no index is returned from the loop, we reach the last line of the function.

Return Value

    Returning p−1p−1: This would return the index 2, which is part of the top 70% of probabilities.
    Returning n−1n−1: This would return the index 3, which is also part of the top 70% of probabilities.

Counter Example

However, consider a slight change in the probabilities:

    probs = [0.4, 0.3, 0.2, 0.1]
    topp = 0.7

In this case:

    probs_sort = [(0.4, 0), (0.3, 1), (0.2, 2), (0.1, 3)]
    The accumulation will again stop at p = 3.

But now:

    Returning p−1p−1: This would return the index 2, part of the top 70%.
    Returning n−1n−1: This would return the index 3, which is not part of the top 70% of probabilities.

Conclusion

The second example demonstrates a situation where returning n−1n−1 would fail to align with the top-p threshold, as it returns an index not part of the top-p probabilities. By returning p−1p−1, the code would ensure that the return value is consistent with the top-p threshold, even in this specific case.

Thank you LLMs for that succinct appraisal, what tring said.

run.c
bjmsong: yeah, you are right, I have updated it.

twobob (changes requested):
I second @tringwald for p -1

twobob (commented):
seems to fix it. thanks

run.c
bjmsong: Thank you both for advice. fixed

twobob (commented):
My only concern is that the obfuscates away the pedagogical nature of the simplistic core sampling

Perhaps consider including both and setting top_p to 1.0 by default and only using the top_p tranche when lower than 1.0?

`
typedef struct {
    float value;
    int index;
} IndexedFloat;

int compare_indexed_float(const void* a, const void* b) {
    IndexedFloat* indexed_a = (IndexedFloat*)a;
    IndexedFloat* indexed_b = (IndexedFloat*)b;
    return (indexed_b->value > indexed_a->value) ? 1 : ((indexed_b->value < indexed_a->value) ? -1 : 0);
}

int sample_top_p(float* probs, int n, float topp) {
    IndexedFloat probs_sort[n];
    for (int i = 0; i < n; i++) {
        probs_sort[i].value = probs[i];
        probs_sort[i].index = i;
    }
    qsort(probs_sort, n, sizeof(IndexedFloat), compare_indexed_float);

    float accum = 0.f;
    int p = 0;
    for (; accum<=topp && p < n; p++) {
        accum += probs_sort[p].value;
    }

    float r = random_f32() * accum;
    float cdf = 0.0f;
    for (int i = 0; i < p; i++) {
        cdf += probs_sort[i].value;
        if (r < cdf) {
            return probs_sort[i].index;
        }
    }
    return probs_sort[p-1].index; // in case of rounding errors
}



int sample(float* probabilities, int n) {
    // sample index from probabilities, they must sum to 1
    float r = random_f32();
    float cdf = 0.0f;
    for (int i = 0; i < n; i++) {
        cdf += probabilities[i];
        if (r < cdf) {
            return i;
        }
    }
    return n - 1; // in case of rounding errors
}
`

and then
`
	float top_p = 1.0f;	
	...snip...
	        top_p = atof(argv[4]);
	        ...snip...
				if ( top_p < 1.0)
				next = sample_top_p(state.logits, config.vocab_size, top_p);
				else			
				next = sample(state.logits, config.vocab_size);`

or similar?

such an approach is not sigficantly longer than eliding the current sampling and I suspect it will also appease those who don't want the overhead?

twobob:
for /f "tokens=4" %i in ('chcp') do @(chcp 65001 && @echo off && for /L %j in (1,1,3) do @(for %x in (runmingw.exe rungcc.exe run.exe) do @(set OMP_NUM_THREADS=4 && %x ../out/model110M.bin 0 0 .99999 "test sampling" 0)) && @echo on && chcp %i)

achieved tok/s: 63.320129 for MINGW
achieved tok/s: 65.864023 for GCC
achieved tok/s: 75.082569 for CLANG
(precis of best of 3)

Also tested the top_p=1.0 workflow just for a "control" on my hardware, that just uses sample()
    ... @(set OMP_NUM_THREADS=4 && %x ../out/model110M.bin 0 0 1 "test sampling" 0)...

achieved tok/s: 65.209077 for MINGW
achieved tok/s: 68.200000 for GCC
achieved tok/s: 76.394593 for CLANG
(precis of best of 3)

compares favourably for the additional funcitonality, control numbers makes sense. 
works on various compilers on windows.

twobob (approved):

karpathy:
I added my own version here because I wanted to change a bunch of things. Thank you for inspiration on this https://github.com/karpathy/llama2.c/commit/8931d5092e3f3266311722f15601892d5beea394

PR #215, opened 2023-08-02
replaced __int64 with int64_t and DWORD with uint_32

Majdoddin:
following merged RP #179, just replaced `__int64` and `DWORD` with standard C types `int64_t` and `uint_32` respectively. Functionality does not change. 

references:  
https://en.cppreference.com/w/c/types/integer
https://learn.microsoft.com/en-us/cpp/c-runtime-library/standard-types?view=msvc-170

### DIFF
--- a/win.c
+++ b/win.c
@@ -6,7 +6,7 @@
 #define FILE_MAP_EXECUTE    0x0020
 #endif /* FILE_MAP_EXECUTE */
 
-static int __map_mman_error(const DWORD err, const int deferr)
+static int __map_mman_error(const uint32_t err, const int deferr)
 {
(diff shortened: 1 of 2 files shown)


karpathy:
@richinseattle please have a look

richinseattle:
@karpathy this is fine to merge.  

This is just a stylistic patch. Personally, I do prefer explicit stdint.h types over the Windows convention of using DWORD and we've already run into the inconsistencies of types across platforms with `long` so this is fine for clarity, especially for primarily Linux developers unfamiliar with Windows types. I'd like to see the switch to stdint in run.c as well in a separate patch.

PR #218, opened 2023-08-02
Proper cmd line argument parsing

tring27:
Not so poor anymore, adding some homegrown command line argument parsing to the c code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,14 @@ Compile and run the C code:
 
 ```bash
 make run
-./run stories15M.bin
+./run -cp stories15M.bin
 ```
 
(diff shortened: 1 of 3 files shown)


karpathy:
So actually I like something like this, but I want the checkoint to be required positional first arg. The rest of them to be optional kwargs

tring27:
With this PR, checkpoint is a required arg, rest all are optional.
Its just that the position of any args are flexible. So as long as user provides a `-cp` the executable will run, otherwise it will print the usage menu.

trholding:
Oops, I didn't see this one. 

Here is an alternate approach which I implemented in my fork.

https://github.com/karpathy/llama2.c/pull/222

karpathy:
Closing this one but looking at #222

PR #219, opened 2023-08-02
Simplify main loop logic

ozabluda:
Simplify main loop logic by converting into a for loop


karpathy:
a little too crowded but ty

PR #222, opened 2023-08-03
run.c - Enhance arg parse : optional parameters

trholding:
It would be ideal to specify parameters without having to set all. One could use getopt for that but it is not portable. So we roll our own minimal one.

Usage:

```bash
run <checkpoint_file> -t [temperature] -s [steps] -p [prompt]
```

Minimum:

```bash
run <checkpoint_file>
```

The following are optional:

-t [temperature] -s [steps] -p [prompt]

if they are omitted, default values are used.

The order of -t -s -p does not matter but the checkpoint_file has to be the first parameter.

To keep it simple without needing for extended checks, if the parameter to the last option is missing, the current default value is used instead.

Updated README.MD to reflect changes.

Incorporated requirement from: https://github.com/karpathy/llama2.c/pull/218#issuecomment-1662596666

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,24 +20,31 @@ Compile and run the C code:
 
 ```bash
 make run
-./run stories15M.bin
+./run -c stories15M.bin
 ```
 
 You'll see the text stream a sample. On my M1 MacBook Air this runs at ~110 tokens/s. See [performance](#performance) or the Makefile for compile flags that can significantly speed this up. We can also try a bit bigger 42M parameter model:
 
 ```bash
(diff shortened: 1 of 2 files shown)


trholding:
TODO:
Wait for status on https://github.com/karpathy/llama2.c/pull/218 
If required add https://github.com/karpathy/llama2.c/pull/218#issuecomment-1662596666

karpathy:
Ok I wrote my own here: 

https://github.com/karpathy/llama2.c/commit/dcef5ff7c720265752b7963c244c70846d6be252

Thank you for the inspiration!

PR #224, opened 2023-08-03
Generic data loader *.py function to load any dataset with any tokenizer

LexiestLeszek:
# Generic dataset loader: use it to download and pretokenize any dataset from Hugging Face

# To download the dataset use the name:
#    $: python load_data.py download --dataset_name roneneldan/TinyStories

# To pretokenize the dataset using the GPT-2 tokenizer:
#    $: python load_data.py pretokenize --dataset_name roneneldan/TinyStories --tokenizer_name gpt2 --max_seq_len 512

### DIFF
--- a/load_data.py
+++ b/load_data.py
@@ -1,0 +1,133 @@
+# Generic dataset loader: use it to download and pretokenize any dataset from Hugging Face
+# Made by Leszek Mielnikow
+
+# To download the dataset use the name:
+#    $: python load_data.py download --dataset_name roneneldan/TinyStories
+
+# To pretokenize the dataset using the GPT-2 tokenizer:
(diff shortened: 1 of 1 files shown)


karpathy:
you wouldn't want to tokenize with gpt-2 tokenizer, everything would break. have to use the llama 2 tokenizer. all that said i'm not sure i want to merge this to master.

LexiestLeszek:
i wanted to use my own dataset based on some standup comedian's jokes concatenated specials' text and that's how I come up with the idea to make a generic dataset loader.

Do you think it would be useful to write only a dataset loader while keeping the original tokenizer unchanged?

karpathy:
I mean yes the script works with any tokenizer, but this repo is for training and running Llama 2, so it feels weird to use inconsistent tokenizers.

Issue #209, opened 2023-08-01
prompt doesn't handle all characters


stellanhaglund:
when I prompt with characters like ö, I get not good as the response.

karpathy:
There might be a problem handling non-ASCII characters in the prompt. I'm not sure how C handles this. Python version uses utf-8 encoding.

twobob:
**Assuming windows** 
The command chcp 65001 is used to change the active console code page to UTF-8 on Windows
you could try opening **cmd** and typing

> **chcp** 
> to get your current codepage  - my default is 437
> if you wish to reset your codepage in the terminal at any time you can then just run **chcp 437** (but your number) to do so.

The following will attempt to automatically grab and restore your current codepage gracefully (see above for any issues) and swap it to utf-8 before you run your command
**for /f "tokens=4" %i in ('chcp') do (chcp 65001 && run.exe ./out/stories44m.bin 0 0 "My utf8 prömpt" && chcp %i)**

no way to test if that is the issue here but that would be the thrust of it.
assuming that was the issue you could elect to either run a term in chcp 65001 or do the kind of switcheroo wangle outlined in the oneliner

Hope it helps

richinseattle:
On Windows, it can be as easy as using `wmain(int argc, char **wchar_t argv)` instead of traditional main to support unicode strings, but that differs from the solution on Linux.

atamurad:
I checked what's going wrong here.

Issue 1: Need to merge multi-byte UTF8 bytes when doing initial str_lookup in bpe_encode.
Smth like this works, not sure if there's a simpler way:
```C
    for (char *c = text; *c != '\0'; c++) {

        // check if char is part of multi-byte UTF8 encoding
        if( (*c & 0xc0) == 0x80) {
            // append it to buffer
            size_t str_len = strlen(str_buffer);
            str_buffer[str_len] = *c;
            str_buffer[str_len+1] = '\0';
        } else {
            sprintf(str_buffer, "%c", *c);
        }

        int id = str_lookup(str_buffer, vocab, vocab_size);

        if(id != -1) {
            tokens[*n_tokens] = id;
            (*n_tokens)++;
        }
    }

```

[...]

Issue #204, opened 2023-07-31
Question - Weird prompt results


stellanhaglund:
I have trained a model with `max_seq_len=512` and examples in this format `some-text -> some-answer###` 
when I run the model without prompt i get new random data in the correct format but when I start with the prompt 
`"some-text ->" `I get only weird results and usually another -> with a good answer after it.

I'm not sure this is because of the prompting but it's strange that the outputs without a prompt works a lot better than with. 

I thought it could be because the examples where of different lengths so I tried padding them, but the tokenizer has -1 for pad_id so tried padding with the eos_id instead, but there doesn't seem to be any difference.

Does anyone have any insights here?

stellanhaglund:
The model has a very good validation loss after 100k iterations, 0.35 so I guess it should be working good..

sri9s:
Hi @stellanhaglund, do you actually have the < symbol in your dataset in the format you specified? What is your dataset size?

stellanhaglund:
Yes I have it in the dataset.
It’s 900k rows with 250-500 tokens per row but padded to 512 tokens per row.

stellanhaglund:
I tried most of the saved models along the training and its the same problem on all of them, so I don’t think its overfitted

stellanhaglund:
from what I can tell the token encoder in run.c gives a result quite different from the python version I will try to hardcode the encoded prompt from the python version and see if I get better results

stellanhaglund:
Ok so that was the issue, hardcoding the encoded prompt from the python version gives me perfect output, @karpathy do you have any ideas where the encoding differs from the original sentencepiece encoding?

karpathy:
:| oh no... not really sadly, if someone has some cycles to track this down would welcome a PR fix

stellanhaglund:
I tried adding sentencepiece in rust to llama2.rs and the prompting works good there, though a bit slower. 
Looked a bit at the c++ implementation and it doesn't look overly complicated, though it uses a hash-map, could probably be worked around with some looping though, but maybe not as performant. 

https://github.com/google/sentencepiece/blob/635fe8423a249b6e081aacd290d8aef7476c6a28/src/bpe_model.cc

twobob:
the C++ version includes additional features like BPE-dropout and resegmentation over the current. I didn't find the python version on first glance? but perhaps it's some detail like that.

stellanhaglund:
[...]

PR #226, opened 2023-08-04
Prompt tokenizer improvements (UTF-8 support, speed-up & compatibilities)

atamurad:
## UTF-8 support

**Test 1: "Lets try ö & 株式会社"**

sentencepiece output:
```
In [53]: s.encode("Lets try ö & 株式会社")
Out[53]: [365, 1691, 1018, 3963, 669, 29871, 31409, 30607, 30437, 30564]
```

run.c debug output:
```
365, 1691, 1018, 3963, 669, 29871, 31409, 30607, 30437, 30564,
```

## Byte Fallback

This PR also adds default `add_dummy_prefix` and `byte_fallback` options compatibility with sentencepiece.

**Test 2: Byte Fallback, Unicode character not in the vocabulary**
```
In [51]: runc_tokens = [29871, 243, 149, 145, 154, 243, 150, 147, 144, ]

In [52]: s.encode("𒎗𓐍") == runc_tokens
Out[52]: True

```
## Speed-up

https://github.com/karpathy/llama2.c/pull/251 includes sorted vocabulary + binary search for faster lookup in vocabulary, among other optimizations. I've updated this PR to use sorted vocabulary with `qsort` and `bsearch`, both standard C library functions. 

**Test 3: Long prompt - 256 word prompt from `lorem -n 256`**

| branch | bpe_encode run time |
|---|----|
| master | 56.89 secs. |
| this PR / sorted vocabulary |  0.30 secs. |
| https://github.com/karpathy/llama2.c/pull/251 | 0.01 secs. |

## One char tokens bugfix

**Test 4: New tok512 model, prompt 'abcdefgh'**

`tokenizer.py` converts ''<0x01>' type tokens into '\x01' values, but these tokens are reserved for byte fallback encoding, and not used for encoding normal characters in string.

```
In [12]: s = spm.SentencePieceProcessor(model_file='tok512.model')

In [13]: s.encode("abcdefgh")
Out[13]: [261, 430, 429, 418, 411, 431, 428, 415]
```

And here is the completely different output from the current code:
```
100 a
101 b
102 c
103 d
104 e
105 f
106 g
107 h
```

Correct output from fixed version:
```
./run out/stories260K.bin -z tok512.bin -i "abcdefgh"
261  a
430 b
429 c
418 d
411 e
431 f
428 g
415 h
```

---

Fixes https://github.com/karpathy/llama2.c/issues/209, https://github.com/karpathy/llama2.c/issues/204


xefoci7612:
This does not seem to cover all UTF-8 cases. I'd suggest using something like this, eventually in a `while `loop:

```C

char* next_utf8(char* text) {

    if (*text == '\0')
        return text;

    size_t len = 0;
    unsigned char lb = (unsigned char)(*text);

    if ((lb & 0x80) == 0)         // lead bit is zero, must be a single ascii
       len = 1;
    else if ((lb & 0xE0) == 0xC0) // 110x xxxx
       len = 2;
    else if ((lb & 0xF0) == 0xE0) // 1110 xxxx
       len = 3;
    else if ((lb & 0xF8) == 0xF0) // 1111 0xxx
       len = 4;
    else {
       printf("Unrecognized lead byte (%s)\n", text);
       exit(1);
    }    
    return text + len;
}

```

atamurad:
@xefoci7612 it does cover all valid utf8 encodings. As you can see from the test case there are 2-byte and 3-byte characters that are encoded correctly. 

TODO now is how to handle multi-byte characters that are not in the vocabulary (fall back to byte-level tokens) and handle invalid encodings/byte streams (fail with error message).

atamurad:
sentencepiece encodes invalid utf8 byte sequences as byte tokens, so now only need to implement byte-level fallback if vocabulary lookup fails.

```
In [30]: invalid = b"\xe4\xba\xbb\xbc\xbd"

In [31]: s.encode(invalid)
Out[31]: [29871, 231, 189, 190, 26308]

In [32]: invalid.decode("utf8")
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
Cell In[32], line 1
----> 1 invalid.decode("utf8")

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbc in position 3: invalid start byte

```

xefoci7612:
> @xefoci7612 it does cover all valid utf8 encodings. As you can see from the test case there are 2-byte and 3-byte characters that are encoded correctly.
> 

Your code relies on the fact that multi byte chars do not overlap:

Suppose you have 2 characters A with 2 bytes and B with 3 bytes where

```
A = 110xxxxx 10xxxxxx
B = 1110xxxx 10xxxxxx 10xxxxxx
```
If second byte of A == third byte of B and we have B in the prompt, then token of A will be added instead of B.

The bug is you call str_lookup() at every step, even in the middle of a multi-byte char

```C
    for (char *c = text; *c != '\0'; c++) {

        // check if char is part of multi-byte UTF8 encoding
        if( (*c & 0xc0) == 0x80) {
            // append it to buffer
            size_t str_len = strlen(str_buffer);
            str_buffer[str_len] = *c;
            str_buffer[str_len+1] = '\0';
        } else {
            sprintf(str_buffer, "%c", *c);
        }

        int id = str_lookup(str_buffer, vocab, vocab_size);

```

atamurad:
> Your code relies on the fact that multi byte chars do not overlap:

They do not overlap because leading bits differ for 2-byte / 3-byte / 4-byte encodings.

Note the different leading bits in your example:
> If second byte of A == third byte of B 
```
A =          [110]xxxxx 10xxxxxx
B = 1110xxxx [10]xxxxxx 10xxxxxx
```

atamurad:
> It is if last byte of A == second byte of B

Leading bits are still different in this case too, so there is no issue unless there is an invalid utf8 sequence in the vocabulary:

```
A = [110]xxxxx 10xxxxxx
B = [1110]xxxx 10xxxxxx 10xxxxxx
```

xefoci7612:
> > It is if last byte of A == second byte of B
> 
> Leading bits are still different in this case too, so there is no issue unless there is an invalid utf8 sequence in the vocabulary:
> 
> ```
> A = [110]xxxxx 10xxxxxx
> B = [1110]xxxx 10xxxxxx 10xxxxxx
> ```

Ok, now I see your point. You are right.

xefoci7612:
> > You keep calling str_lookup() until it succeeds!!! Well that's quite tricky IMHO, but of course it is subjective.
> 
> This checks if the current byte is the last byte of multi-byte sequence by peeking next character:
> 
> ```
> if( (*(c+1) & 0xC0) != 0x80) {
>     // this is the last byte of multi-byte sequence and not in vocabulary
> ```
> 
> We can add the same check before calling str_lookup but I am not sure it's worth it.

Untested.

```C
    char* e = str_buffer;
    do
        *e++ = *c++;
    while ( (*c & 0xc0) == 0x80);
    *e = '\0';
```

Note that first char is always copied.

iSevenDays:
Thanks for the PR! Now I can test other languages other than English

twobob:
This fixes #204 , it has been reported.

karpathy:
@atamurad can we not sort the vocab in python so we don't have to have all these new lines of code in run.c?

atamurad:
> @atamurad can we not sort the vocab in python so we don't have to have all these new lines of code in run.c?

@karpathy I tried this now as a draft but some lines are required to alloc/parse/free the sorted order, and we need to keep the original order/permutation too (needed for printing tokens). 

It introduces two additional arrays, and I think net LOC is probably the same if not more..

sorted_vocab
scores
sorted_idx_to_ids // map from sorted to token ID
id_to_sorted_idx // map from token ID to sorted vocab idx


Also forks in other languages parse tokenizer.bin too, so probably should try to keep the structure as simple as possible and not change it too much IMO?

Another approach I can look into is hashing and/or maybe collapse/fold `compare_tokens`, `str_lookup` and `TokenIndex` to one liner.

karpathy:
Got it ok. I took a good chunk of time yesterday to review the UTF-8 encoding details and this PR's implementation. I think most people won't be familiar with UTF-8 encoding details, and I don't want them to be too scared about these parts, so I think it's important for this to be as clean, simple, localized and educational (in comments) as possible. I'll continue after work today.

atamurad:
@karpathy  Testing of run.c functions from pytest could look like this: 

https://github.com/atamurad/llama2.c/commit/a4ecf6c09fc0aea84dd2c6d7e166d24818d41879

Added new target to Makefile to compile run.c as shared object, which can be loaded in Python with ctypes, and C functions are called directly from python, without modifying anything in run.c.

Added 2 tests to test_all.py

test_load_tokenizer - Loads tok512.bin with run.c, tok512.model with sentencepiece and compare all tokens and scores
test_bpe_encode - Test run.c bpe_encode() output with expected spm.encode() output for given test prompts

Test fails now because it downloads tok512.bin from HF but it should work/pass after HF models are updated.
```
--------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------
Testing bpe_encode with prompt 'hello world!'
Expected: [281, 306, 414, 263, 304, 341, 443]
=============================================================================== short test summary info ================================================================================
FAILED test_all.py::test_load_tokenizer - AssertionError: assert b'' == b'<0x00>'
FAILED test_all.py::test_bpe_encode - assert 36 == 443
```

Let me know what you think, I can update this PR or submit a new one.

run.c
```diff
@@ -399,6 +432,20 @@ void bpe_encode(char *text, char **vocab, float *vocab_scores, int vocab_size, u
...
+    free(sorted_vocab);
+}
+
+// convert token to printable string
+char *token_to_str(char **vocab, int token, int prev_token) {
+    // following BOS (1) token, sentencepiece decoder strips any leading whitespace (see PR #89)
+    char *token_str = (prev_token == 1 && vocab[token][0] == ' ') ? vocab[token]+1 : vocab[token];
+    // make '<0x01>' into '\x01'
+    static char byte_piece[4];
+    if (sscanf(token_str, "<0x%02X>", (int*)(&byte_piece)) == 1) {
```
karpathy, line 444: is this part necessary? maybe we could just leave the strings like <0x01> as they are?

atamurad: Depends on basic alphabet coverage in vocabulary. I can check llama2 tokenizer vocabulary for letters/languages I know but can't comment for other languages. 

if we remove those, we can also remove byte_fallback encoding too, and error out if token is not in vocabulary as before?
```
            // byte_fallback encoding
            for (int i=0; i<str_len; i++) {
                tokens[(*n_tokens)++] = (unsigned char)str_buffer[i] + 3;
            }
```

atamurad: if i remember correctly sentencepiece maintains decode(encode(normalize(s))) == normalize(s) property. decode() will output control characters as is if it was in original string/token.

karpathy:
Also I noticed that the previous code did not actually respect `add_dummy_prefix = True`. Was this a bug

run.c
atamurad: also there is iscntrl() C function but looks like '\n' is there too, and we probably want to print newlines as is.

atamurad:
> Also I noticed that the previous code did not actually respect `add_dummy_prefix = True`. Was this a bug

yes it was. I don't think it is a big issue but good to keep compatibility.

run.c
atamurad: update: 10 languages I tested from lorem ipsum text generator (https://generator.lorem-ipsum.info/), only 3 did not require byte_fallback when encoding 1st short paragraph.

clebert:
I am unable to successfully reproduce the "One char tokens bugfix" test case with the current master version and the following command: `make runfast && ./run stories260K.bin -z tok512.bin -i "abcdefgh"`. Instead of the expected sequence `261, 430, 429, 418, 411, 431, 428, 415`, I get `261 101 102 103 411 105 106 107`.

I observe the same phenomenon with my [Zig Port](https://github.com/clebert/llama2.zig/commit/13e266077f548a05d76aba154b68c8f23845b03c) where I get the sequence `261 430 429 103 104 431, 428, 415`.

atamurad:
@clebert Please run `python3 tokenizer.py -t tok512.model` to re-export tok512.bin (these files are out of repo) and try again. It is very likely you have tok512.bin exported with tokenizer.py before bugfix.

PR #230, opened 2023-08-04
Make pytorch model compatible with TorchScript/torch.jit.save

mpcusack:
Changes a few bits of syntax that the TorchScript compiler doesn't support:
- vararg expansion in `view` and `reshape`. Both of these support being passed a list instead.
- Mixing `int` and `float` in a ternary
- TorchScript assumes that the module's `forward` function will return a single `Tensor` or it raises a `Expected Tensor but got Tuple` error on deserialization. To accomplish this, but otherwise make minimal changes to the structure of the code I assign the loss result to an attribute which can be accessed from the calling context.

Also added a small python script for demonstrating how to save the model.


mpcusack:
Unrelated and would do it in seperate PR, but would you consider a `.gitignore` or something like:
```
data/
__pycache__/
run
stories*.bin
*.pt
```

karpathy:
ok i see, i'll take it thanks

PR #240, opened 2023-08-06
Added support for OpenBLAS

rdentato:
OpenBLAS performs slightly worse than openmp on my machine. 
Nevertheless is around a 5x improvement over the plain CPU run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,10 @@ rungnu:
 runompgnu:
 	$(CC) -Ofast -fopenmp -std=gnu11 run.c  -lm  -o run
 
+.PHONY: runblas
+runblas: run.c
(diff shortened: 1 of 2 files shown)


karpathy:
If OpenBLAS is worse then why add it

xefoci7612:
> OpenBLAS performs slightly worse than openmp on my machine. Nevertheless is around a 5x improvement over the plain CPU run.

@rdentato can you make it work in multi-thread? Does the cblas_sgemv() call start parallel threads to do the work? If not, I'm afraid it's a hard sell...

I don't know BLAS, so I can't help on this one.

rdentato:
I think this is worthwile to investigate further if anyone confirms that there are cases where openmp is not available and openBLAS is. 
Otherwise, I would just rely on OpenMP and discard the PR.

karpathy:
no gain no pain

PR #243, opened 2023-08-06
Stop the text generation if a BOS is generated

rdentato:
If a BOS is generated, it means that the text generation is following a path that is not related to the input prompt.
I believe this is useless and would be better to stop the text generation altogether.
Also, since the generation can stop earlier than `steps`  steps, this is to be taken into account when calculating the tok/sec achieved.

### DIFF
--- a/run.c
+++ b/run.c
@@ -632,6 +632,8 @@ int main(int argc, char *argv[]) {
                 }
             }
         }
+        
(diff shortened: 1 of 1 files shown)


karpathy:
Added here ok sounds good https://github.com/karpathy/llama2.c/commit/79791f39b49703f14fb015b558a2d8d6e692eb49

PR #247, opened 2023-08-06
Speed up streaming output by buffering till newline

ozabluda:
Inspired by https://github.com/karpathy/llama2.c/pull/193, see that for details.

Performance difference between buffering beyond newline isn't measurable, but this is much simpler. Current implementation only supports 0 and 1 tokens, again for simplicity. Buffering all output can be achieved by redirecting to a file. Unclear if the number of tokens other than 0, 1, inf is useful.


karpathy:
Is this even faster / does it matter? I don't really think so?

ozabluda:
In the PR this was inspired by people report up to 50-80% speedups. I get a couple of percent speedup. Presumably the effect is larger on fast machines (GPUs/etc) with small models.

PR #248, opened 2023-08-06
Implement token healing

xefoci7612:
Fix the so called "mid-token" issue by stripping the last token from the prompt and then sampling the first processed token after the prompt among the ones that start with the last token's string.

This is compatible with any kind of sampling and is called "token healing": https://github.com/microsoft/guidance/blob/main/notebooks/token_healing.ipynb

We also got a new feature for free. We can now explicitly set last word boundary in the prompt, i.e. signal if the last word of the prompt should be considered a full word or just a partial word.

For instance, in this case:

```
./run model.bin 0 5 "On"
<s>
One day, a little
```

Last word of the prompt is expanded On -> One by the model. Instead if we want the sentence to specifically start with the word "On" we can now simply submit:

```
./run model.bin 0 5 "On "
<s>
On a hot summer day
```

### DIFF
--- a/run.c
+++ b/run.c
@@ -495,6 +495,15 @@ int sample_topp(float* probabilities, int n, float topp, ProbIndex* probindex) {
     return probindex[last_idx].index; // in case of rounding errors
 }
 
+void clamp_logits(float* logits, int last_token, char** vocab, int vocab_size) {
+    char* last = vocab[last_token];
+    size_t len = strlen(last);
+    for (int i = 0; i < vocab_size; i++) {
+        if (strncmp(last, vocab[i], len) != 0) {
+            logits[i] = -1e10;
+        }
+    }
+}
 
 // ----------------------------------------------------------------------------
 // int main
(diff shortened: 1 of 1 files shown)


xefoci7612:
This new version replaces the old one  #191

karpathy:
Sadly token healing is a lot more complex issue than the microsoft document lets on. Actually their "solution" is too naive and wrong. The problem is that both 1) token boundaries can be "unstable" as you add tokens, with boundaries potentially moving around, and 2) you might need multiple tokens. The fully correct solution is computationally and algorithmically quite complex - you have to back up potentially multiple tokens, then keep track of a sampling tree and eliminate prompt-inconsistent branches. One implementation seems to be hinted at deep inside [tiktoken](https://github.com/openai/tiktoken/blob/5d970c1100d3210b42497203d6b5c1e30cfda6cb/src/lib.rs#L318), it's not even documented. Basically it's up to the user to end their prompts at "stable" and reliable token boundaries, e.g. usually right before a space. Then you're not in trouble. Also note that Chat models don't suffer from this too much because there is no prompting in a sense of contrained prefix sampling for the Assistant responses.

So yes this is one great reason to dislike tokenization and cheer for the byte-level Transformer papers.

PR #249, opened 2023-08-06
Added conditional include of the OpenMP header.

rdentato:
Since the OMP pragmas are used, it seems appropriate to include the openMP header (if the -fopenmp flag is specified).

### DIFF
--- a/run.c
+++ b/run.c
@@ -20,6 +20,11 @@ $ ./run
     #include <unistd.h>
     #include <sys/mman.h>
 #endif
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
 // ----------------------------------------------------------------------------
 // Transformer and RunState structs, and related memory management
 


twobob (commented):
Confirmed. this is how I run it anyway. But Andrej said somewhere he didn't like #ifdef

so I didn't suggest this. Just letting you know

karpathy:
This makes sense to me... actually wait why does OpenMP work on my Linux system without this import?

rdentato:
It currently works because only #pragma is used.
I tried to use the `omp_get_num_devices()` and noticed the header was missing.

twobob:
> It currently works because only #pragma is used. I tried to use the `omp_get_num_devices()` and noticed the header was missing. So, it's not strictly needed but you never know if some compiler uses it or not (I use gcc 11 but someone might use gcc 12 or clang).

I use clang, gcc and mingw to side-by-side perf them.
so again. confirmed that someone does that (and why I already roll like that)

PR #250, opened 2023-08-06
FIX: model.generate(); forward() only returns logits now.

npinto:
This patch fixes a simple bug in `generate()` due to model's `forward()` only returning logits and not losses since `f2e34e6b0ac55accd6ba930a04c6f683f5158b29`.


karpathy:
oops ty

PR #251, opened 2023-08-07
Faster tokenization (log(n) lookups & caching of possible merges)

ochafik:
This greatly speeds up tokenization of large prompts (up to 3000x faster for 2k prompt tokens), making it possible to use prompts of the full context length without paying quadratic costs.

Tested `runfast` builds on an M2 Max Mac w/ [hyperfine](https://github.com/sharkdp/hyperfine) & [lorem](https://formulae.brew.sh/formula/lorem):
```bash
# `lorem -n 128` yields 336 tokens;
# Only generating / consuming one token to test tokenization itself
hyperfine -L bin run_fasttoken,run_master --warmup=1 "./{bin} llama2.c.stories110M.bin -n 1 -i \"\$( lorem -n 128 )\""
```

Result:
```
Benchmark 1: ./run_fasttoken "llama2.c.stories110M.bin" -n 1 -i "$( lorem -n 128 )"
  Time (mean ± σ):      86.7 ms ±   2.9 ms    [User: 32.5 ms, System: 31.0 ms]
  Range (min … max):    81.8 ms …  90.6 ms    10 runs
 
Benchmark 2: ./run_master "llama2.c.stories110M.bin" -n 1 -i "$( lorem -n 128 )"
  Time (mean ± σ):     11.151 s ±  0.177 s    [User: 10.927 s, System: 0.168 s]
  Range (min … max):   10.950 s … 11.563 s    10 runs
 
Summary
  './run_fasttoken "llama2.c.stories110M.bin" -n 1 -i "$( lorem -n 128 )"' ran
  128.61 ± 4.72 times faster than './run_master "llama2.c.stories110M.bin" -n 1 -i "$( lorem -n 128 )"'
```

With `lorem -n 750` (1997 tokens), this PR's build takes 100ms and master takes ~6min.

PR #253, opened 2023-08-07
Faster prompt eval w/ early exit after last layer's kv cache write

ochafik:
Prompt evaluation's only side effect is to build the kv cache for all the layers, so we can exit early at the last layer once we've cached its k & v vectors (and also skip the final rmsnorm & logits computation)

Tested: `make runfast` on M2 Max Mac and adjusted `-n` to only test prompt eval using prompts generated w/ [lorem](https://formulae.brew.sh/formula/lorem):

```bash
./run llama2.c.stories110M.bin -n 336 -i "$( lorem -n 128 )"
./run llama2.c.stories110M.bin -n 1997 -i "$( lorem -n 750 )"
```

| -n | master | this PR | speedup wrt/ master | fast tok (https://github.com/karpathy/llama2.c/pull/251) | this PR + fast tok | speedup wrt/ fast tok |
| -- |  -- |  -- |  -- |  -- |  -- |  -- | 
| 336 | 14.8sec | 13.6sec | -8% | 3.5sec | 2.5sec | -28% (?) |
| 1997 | - | - | - | 38.6sec | 32.2sec | -16% |  

Notes:
- Tokenization is very slow in `master` and dominates the execution time. I've added figures after merging the fast tokenizer from https://github.com/karpathy/llama2.c/pull/251 and using it as baseline
- Some numbers are a bit surprising (excessive speed up), I'll test again when I get a chance.

### DIFF
--- a/run.c
+++ b/run.c
@@ -10,6 +10,7 @@ $ ./run
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <time.h>
 #include <math.h>
 #include <string.h>
@@ -213,7 +214,7 @@ void matmul(float* xout, float* x, float* w, int n, int d) {
     }
 }
 
-void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights* w) {
+void transformer(int token, int pos, Config* p, RunState* s, TransformerWeights* w, bool evaluating_prompt) {
 
     // a few convenience variables
     float *x = s->x;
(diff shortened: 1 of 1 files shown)


karpathy:
not willing to bloat things just yet for this speedup

ochafik:
@karpathy thanks for taking a look (and for the cool repo!). Note that after https://github.com/karpathy/llama2.c/pull/251 it would be a 11-16% speedup for 10 extra LOC (updated figures), but fair enough ✌️

PR #254, opened 2023-08-07
Add -a flag to specify additional steps after prompt

ochafik:
Makes it easy to control how many tokens are generated vs. simply passed through from the prompt (not trivial to guess how many tokens the prompt is split into).


karpathy:
too much complexity introduced imo on this one

PR #256, opened 2023-08-07
Patch rng seed

rdentato:
Rather than exit, let's treat rng_seed = 0 as an indication for a default value (time(NULL)).
It seems more reasonable to me ...

### DIFF
--- a/run.c
+++ b/run.c
@@ -20,6 +20,11 @@ $ ./run
     #include <unistd.h>
     #include <sys/mman.h>
 #endif
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
 // ----------------------------------------------------------------------------
 // Transformer and RunState structs, and related memory management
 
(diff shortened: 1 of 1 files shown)


rdentato:
The issue I see with 
```
rng_seed = atoi(argv[i + 1]) ^ RND_0;
```
is that if `RND_0` is `0x2545F4914F6CDD1Dull` then 
```
./run model.bin -s 2685821657736338717
```
will set `rng_seed` to 0.

Yes, is very unlikely, but still possible.

The reason behind my proposal is that you can imagine `run` called by a script
```
   ./run model.bin -s ${LLM_SEED:-0}
```
If `LLM_SEED` is not set, the default value is give (and it's a value that changes all the time).
I usually prefer to have a value that signifies *default* and `0` as a seed seems a very good candidate to me.

twobob:
> If `LLM_SEED` is not set, the default value is give (and it's a value that changes all the time). I usually prefer to have a value that signifies _default_ and `0` as a seed seems a very good candidate to me.

Exact reason s=0 was immediately tested here, following the change.  Concur, fwiw, with this mindset.
That said, determinism was also our expectation so, unsure.

PR #257, opened 2023-08-07
docs: added run.c inference architecture

nikolaydubina:
Proposing to add architecture diagram of inference for `run.c`.

I am working in my Go port [llama2.go](http://github.com/nikolaydubina/llama2.go) to extend it to support dialog (which requires live modification of token sequence), and I needed to understand how tensors are affected by token sequence (previous tokens, caches) and how heads work.

I picked whiteboard style and http://tldraw.com/ tool for it, because it looks good. But actually would like know how to better visualise this.

Just raising a proposal here with this example! Thank you! 🙇🏻‍♂️


rdentato:
I believe a diagram like this is of great value. 
I'm not sure I can understand it, are you planning to add any note to it?

karpathy:
very cool, but probably not for master :)

PR #261, opened 2023-08-08
randomized model for CI workflow

Majdoddin:
Augments `train.py` to generate a randomized model. And extends the CI workflow to generate and sample the randomized model. The sample of Ubuntu serves as the reference and other samples are checked against it, as suggested by @karpathy in #175.

arm64 (windows) is excluded, because the binaries from the `amd64_arm64` cross-compilation cannot be run on the `amd64` runner. Selfhosting or QEMU does it?

Interestingly the the output tokens do NOT match for windows and ubuntu! This is the comparison:
`ubuntu` vs `amd64` and `amd64-x86` and `mingw`
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,17 +3,19 @@ name: Continuous Integration
 on:
   push:
     branches:
-      - master
-    paths: ['.github/workflows/**', '**/Makefile', '**/*.c', '**/*.h']
+      - master, 
+    paths: ['.github/workflows/**', '**/Makefile', '**/*.c', '**/*.h', '**/*.py']
   pull_request:
     types: [opened, synchronize, reopened]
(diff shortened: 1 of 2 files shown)


Majdoddin:
> Interestingly the the output tokens do NOT match for windows and ubuntu! This is the comparison: `ubuntu` vs `amd64` and `amd64-x86` and `mingw` 
You can inspect it on the failed checks on this PR

Majdoddin:
The output tokens on my personal windows PC is identical to that of ubuntu.

karpathy:
Instead of randomized model I am planning to train a tiny teeny model and will commit it into the repo (as I expect it will be very small). Then I will loop back around to make tests out of it.

Majdoddin:
@karpathy yeah, but you suggested the randomized model, and it took me a few days to implement ...

PR #262, opened 2023-08-08
Clarify we rotate by complex multiplication.

xefoci7612:
Better document what the code does and simplify it.

### DIFF
--- a/run.c
+++ b/run.c
@@ -213,6 +213,13 @@ void matmul(float* xout, float* x, float* w, int n, int d) {
     }
 }
 
+void complexmul(float* a, float* b, float c, float d) {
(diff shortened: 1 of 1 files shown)


rdentato:
I like it.
I would not shorten `freq_cis_real_row` and `freq_cis_imag_row` in `fcr` and `fci`  and I would rename the argument of `complexmul()` to `a_r`, `a_i`, `b_r` and `b_i`  but that's an absolutely minor point.

Overall, I think is much better to explicitly show that this is a complex multiplication as proposed here.

karpathy:
eeeh i'm a bit shrug on this

PR #265, opened 2023-08-09
Only C complient optimisation for windows and mingw

Majdoddin:
The randomized model of RP #261 makes (slightly) different tokens on ubuntu and macOS on one side and Windows and mingw on the other side. 
This is because the builds use non C complaint optimisation. This RP fixes this issue, so the random model has identical output on all systems.
Notice that ubuntu and macOS builds still use non C complient optimisation, which is safer to fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ runomp: run.c
 
 .PHONY: win64
 win64:
-	x86_64-w64-mingw32-gcc -Ofast -D_WIN32 -o run.exe -I. run.c win.c
+	x86_64-w64-mingw32-gcc -O3 -D_WIN32 -o run.exe -I. run.c win.c
 
 # compiles with gnu99 standard flags for amazon linux, coreos, etc. compatibility
 .PHONY: rungnu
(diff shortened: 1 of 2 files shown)


Majdoddin:
See also #268

karpathy:
Is this needed with the new CI

Majdoddin:
@karpathy 
> Is this needed with the new CI

Yes, **IF** it is important that forwarding the model (with fixed random seed and arguments) generates identical tokens, regardless of the setup (OS, architecture, ...). Notice that a different token can derail the following tokens. 
This can be ensured by compiling with C-conformant compiler flags. 

That #261,  using a randomized model, gets different output, but #301 with a tiny model, gets identical output, can be justified by the former having more entropy and thus covering more edge-cases.

Q- If the performance lost is too much, can the code be written in such a way that even `risky` compiler flags produce identical tokens?

karpathy:
The current CI passes without the PR though right?

Majdoddin:
right.

PR #267, opened 2023-08-10
Update readme for openmp on mac

krrishnarraj:


karpathy:
I'm just really unable to get any substantial gains here, even following the new readme. Am I alone?

krrishnarraj:
Then gain was more than 2x on my macbook with openmp enabled.

```
Chip: M1 pro with 10 cpu cores
Memory: 32GB
OS: 13.5
/usr/bin/gcc --version: Apple clang version 14.0.3
/opt/homebrew/opt/llvm/bin/clang --version: Homebrew clang version 16.0.6
```

Launched the binary as: `./run llama2_7b.bin`

| Run | tok/s |
| --- | --- |
| make runfast | 1.79 |
| make runfast CC=/opt/homebrew/opt/llvm/bin/clang | 1.79 |
| make runomp CC=/opt/homebrew/opt/llvm/bin/clang | 4.01 |
| above binary with OMP_NUM_THREADS=4 | 3.96 |

Not setting `OMP_NUM_THREADS` gave the best performance.

For further debugging, let us know your setup

karpathy:
ok let's go with this for now ty

PR #268, opened 2023-08-10
Running the test model in CI Workflow

Majdoddin:
Extends the CI workflow to sample the test model. The sample of Ubuntu serves as the reference and other samples are checked against it, as suggested by @karpathy in https://github.com/karpathy/llama2.c/pull/261.

Interestingly, the CI tests are passed, but the tests fail for a randomized model. This may indicate that a randomized model is a superior test. See #261 and #265 for details.

The test model is currently stories15M.bin. It is a little bigger than the (GitHub's [warning limit](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#file-size-limits)), but it worked.
Alternatively, one can 
-- train a tiny teeny model and use it (just set TEST_MODEL in build.yml), or 
-- use  [Git Large File Storage](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage). @karpathy I can do it if you wish.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,17 +3,20 @@ name: Continuous Integration
 on:
   push:
     branches:
-      - master
-    paths: ['.github/workflows/**', '**/Makefile', '**/*.c', '**/*.h']
+      - master, 
+    paths: ['.github/workflows/**', '**/Makefile', '**/*.c', '**/*.h', '**/*.py']
   pull_request:
     types: [opened, synchronize, reopened]
-    paths: ['**/Makefile', '**/*.c', '**/*.h']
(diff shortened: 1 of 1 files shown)


karpathy:
@Majdoddin see the new commits esp related to the revival of unit tests! I would welcome and merge a PR that adds running  `pytest` and making sure it passes.

karpathy:
(We shouldn't merge the .bin files into the repo. They should be downloaded from huggingface as done in pytest)

karpathy:
Ok I learned a bit about CI / GitHub actions and trying to wing it myself. It works on Ubuntu. Would welcome a followup PR that runs pytest also on other platforms. Closing this one.

PR #272, opened 2023-08-11
Custom Tokenizer training

karpathy:
Add the ability to train a custom sentencepiece tokenizer with a given vocab_size, and pretok with it. some more changes still needed to merge this branch, in train.py and ofc run.c. did this in a sadly bit ugly, but fully backwards compatible way. basically when we use custom tokenizer we create a whole new directory structure for that.

Empirically, when I train vocab_size=2048 on TinyStories, I roughly match the seq_len of average example in TinyStories to the 32,000 vocab tokenizer of Llama 2. This is because the latter has a ton of tokens for various languages and has to do a lot more.

A smaller vocab_size dramatically reduces the number of parameters because the token embedding table is significantly smaller, and the final linear classifier is also significantly smaller, so everything is faster.

So I thought this would be useful to have for custom, narrow domains.


karpathy:
This is WIP btw, putting it here in case people have comments, just to show partial work towards it. But the remaining changes are fairly minimal, in train.py, and in run.c. I suspect this is 90% there.

karpathy:
This makes my `data` directory look as follows for example:

```
(pytorch2) ubuntu:~/llamac$ ls -lhtr data
total 1.5G
-rw-rw-r-- 1 ubuntu ubuntu 1.5G Jul 20 14:47 TinyStories_all_data.tar.gz
drwxr-xr-x 2 ubuntu ubuntu 4.0K Jul 20 16:12 TinyStories_all_data
-rw-rw-r-- 1 ubuntu ubuntu 4.4K Aug 11 03:20 tok512.vocab
-rw-rw-r-- 1 ubuntu ubuntu 7.5K Aug 11 03:20 tok512.model
-rw-rw-r-- 1 ubuntu ubuntu  15K Aug 11 03:24 tok1024.model
-rw-rw-r-- 1 ubuntu ubuntu  11K Aug 11 03:24 tok1024.vocab
-rw-rw-r-- 1 ubuntu ubuntu  24K Aug 11 03:29 tok2048.vocab
-rw-rw-r-- 1 ubuntu ubuntu  31K Aug 11 03:29 tok2048.model
-rw-rw-r-- 1 ubuntu ubuntu  64K Aug 11 03:30 tok4096.model
-rw-rw-r-- 1 ubuntu ubuntu  53K Aug 11 03:30 tok4096.vocab
drwxrwxr-x 2 ubuntu ubuntu 4.0K Aug 11 03:48 tok512
drwxrwxr-x 2 ubuntu ubuntu 4.0K Aug 11 03:50 tok2048
```

where I trained tokenizer using vocab_sizes 512, 1024, 2048, 4096, and then pretokenized 512 and 2048, creating the two directories, which e.g. have: 

```
(pytorch2) ubuntu:~/llamac$ ls -lh data/tok2048
total 2.0G
-rw-rw-r-- 1 ubuntu ubuntu 41M Aug 11 03:50 data00.bin
-rw-rw-r-- 1 ubuntu ubuntu 41M Aug 11 03:50 data01.bin
-rw-rw-r-- 1 ubuntu ubuntu 41M Aug 11 03:50 data02.bin
-rw-rw-r-- 1 ubuntu ubuntu 41M Aug 11 03:50 data03.bin
-rw-rw-r-- 1 ubuntu ubuntu 41M Aug 11 03:50 data04.bin
...
```

(for all 50 shards).

karpathy:
I wish I had thought through the directory structure of a number of potential datasets and a number of potential tokenizers in advance... life

karpathy:
Remaining TODOs:

- model export support
- run.c support
- README docs
- how do we deal with tinyshakespeare? right now it would need large changes to add support. either we do that, or we deprecate it for now. leaning to the latter. the real solution here is a separate PR to make it easier to introduce new datasets with less copy paste code...

PR #274, opened 2023-08-11
Speed up sample_topp

dror-cyata:
This PR should allow using topp logic while avoiding the significant performance hit

Sampling probabilities tend to be sharp, so filtering out low probabilities before sorting the list result with significant time improvement for sample_topp.
We use a threshold that guaranties the filtered entries do not belong to topp (except for maybe a very rare edge case, need to think on it), but anyway we verify it to allow higher threshold.


run.c
```diff
@@ -466,11 +466,29 @@ int sample_topp(float* probabilities, int n, float topp, ProbIndex* probindex) {
...
+        if (probabilities[i] > threshold){
+            ProbIndex t;
+            t = probindex[cur_filtered_index];
+            probindex[i] = probindex[cur_filtered_index];
+            probindex[cur_filtered_index] = t;
+            cur_filtered_index++;
+            cumulative_filtered_prob += probabilities[i];
+        }
         probindex[i].index = i;
         probindex[i].prob = probabilities[i];
```
kroggen, line 488: I don't understand this code. This part appears to be overwriting the above

dror-cyata: You're right. I had a stupid bug. Fixed.

kroggen:
I suspect that #270 is faster

dror-cyata:
> I suspect that #270 is faster

Just looked on #270. Both indeed look similar and filtering out entries with low probabilities, so they should take similar time  (threshold of 1e-5 is used there, and ~3e-6 is used here, by default). However, the threshold used there is arbitrary, and here we used a threshold that assures that the filtered out entries do not belong to topp. Note that in #270 there might be edge cases where entries that are part of topp are unintentionally filtered out.

rdentato:
I agree with @dror-cyata, my constant in PR#270 was arbitrary and I have now replaced it with (1-topp)/(n-1).
The benefit are still substantial. I hope @karpathy will merge one of these PR (there is also #276 on the same topic)

karpathy:
Merged https://github.com/karpathy/llama2.c/pull/276 ty

Issue #246, opened 2023-08-06
Big speed regression with top-p sampling


xefoci7612:
On my system from 40 tok/s down to 33 tok/s, almost 20% slower...

```
$ ./run /tmp/ramdisk/model110m.bin -s 1 -p 0
Once upon a time <stripped> make a big difference in the world.
achieved tok/s: 40.298507
$ ./run /tmp/ramdisk/model110m.bin -s 1
Once upon a time <stripped> make a big difference in the world.
achieved tok/s: 33.588093
```

Slowdown is even more dramatic on smaller models

xefoci7612:
Sorting all the probabilities seems an overkill.

Picking the top K largest values can be done in O(n) by a [selection algorithm](https://en.wikipedia.org/wiki/Selection_algorithm), like  [std::nth_element](https://en.cppreference.com/w/cpp/algorithm/nth_element) in std library.

Eventually we can keep full sorting but after a top-k step. It could be possible to chain top-K and top-P as described [here](https://huggingface.co/blog/how-to-generate):

`While in theory, Top-p seems more elegant than Top-K, both methods work well in practice. Top-p can also be used in combination with Top-K, which can avoid very low ranked words while allowing for some dynamic selection.
`

For instance a top-K with k=60 before top-p will reduce final sorting from 30K elements of vocab to just 60....with a very high probability of getting the same result of a full sort.

twobob:
./run
Usage:   run <checkpoint> [options]
Example: run model.bin -t 0.9 -n 256 -p "Once upon a time"
Options:
  -t <float>  temperature, default 0.9
  -s <int>    random seed, default time(NULL)
  -n <int>    number of steps to run for, default 256. 0 = max_seq_len
  -p <string> prompt string, default none
  -o <float> top_p, default 1.0
  -k <int> print_tokens, 1 or 0 flag default 1
  -f <int> saveFileBool, 1 or 0 flag default 1
  -l <int> saveLogBool, 1 or 0 flag saves timestamps for tokens gen default 0
  -b <int> singleBOS, 1 or 0 flag default 1
  -d <string> dirname string, default none
  
[...]

PR #276, opened 2023-08-12
optimize sample_topp by filtering out small value elements up front

jrudolph:
Refs #246 

This works because we know that in worst case only 1 element will be selected and therefore the remaining (n-1) elements have to split the remaining (1-topp) probability. Probabilities smaller than that cannot be selected and can be filtered out up front.

E.g. for p = 0.9 that means that usually only 100-1000 tokens remain, speeding up the remaining process considerably.

(In llama2.scala, I [further improved](https://github.com/jrudolph/llama2.scala/blob/758de7180b3cafd4b939bf0c16271821d5f46d83/shared/src/main/scala/net/virtualvoid/llama2/Llama2Runner.scala#L42-L51) on that by avoiding the sort in most cases, based on the observation that the distribution looks like power-law and only very few elements will be selected ultimately, so that iteratively scanning over the array to find the next best element (kind of selection sort) keeping track of cumulative p seems to be a slightly better solution yet).

### DIFF
--- a/run.c
+++ b/run.c
@@ -465,17 +465,24 @@ int sample_topp(float* probabilities, int n, float topp, ProbIndex* probindex) {
     // tokens that exceed probability topp. This way we never sample tokens that
     // have very low probabilities and are less likely to go "off the rails".
 
+    int n0 = 0;
     // quicksort indices in descending order of probabilities
+    // elements smaller than (1 - topp) / (n - 1) cannot be part of the result
+    // and can be filtered out directly
+    const float cutoff = (1.0f - topp) / (n - 1);
     for (int i = 0; i < n; i++) {
-        probindex[i].index = i;
-        probindex[i].prob = probabilities[i];
+        if (probabilities[i] >= cutoff) {
+            probindex[n0].index = i;
+            probindex[n0].prob = probabilities[i];
+            n0++;
+        }
     }
-    qsort(probindex, n, sizeof(ProbIndex), compare);
+    qsort(probindex, n0, sizeof(ProbIndex), compare);
 
     // truncate the list where cumulative probability exceeds topp
     float cumulative_prob = 0.0f;
-    int last_idx = 0;
-    for (int i = 0; i < n; i++) {
+    int last_idx = n0 - 1; // in case of rounding errors consider all elements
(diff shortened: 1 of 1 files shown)


cgbur:
238 -> 509 tokens/s when sampling with temperature 1 and top-p 0.9. Really nice work! Did you notice much improvement with the linear scanning? I have not tried that yet in my port. I found the average number of logits per step is pretty drastically reduced already.

image

jrudolph:
Ah, didn't see #270 and #274 came first with the same idea :)

For fun, I ran a code generation model on llama2.scala for a while and asked it for suggestions: https://gist.github.com/jrudolph/fb7641ba2406de705c5499280783b55c

The suggested algorithms are of sometimes comically bad quality but some ideas seem interesting:
 * It suggested QuickSelect (which I think somewhat degenerates for the long-tail distributions we have here, but you could choose clever pivots, basically the filtering in the PR here is just the first step of a QuickSelect)
 * Gathering a histogram in a first run to quickly figure out an upper bound of elements to keep

jrudolph:
> Did you notice much improvement with the linear scanning

I think after the filtering it doesn't matter much any more, after all the speed improvements are only needed anyway for small models (since sampling speed only depends on vocabulary size regardless of model size).

jrudolph:
> I think after the filtering it doesn't matter much any more, after all the speed improvements are only needed anyway for small models (since sampling speed only depends on vocabulary size regardless of model size).

Ok, in Scala, the effect of scanning is improving just the top-p selection process by another 10x (but that's also because the naive idiomatic sorting involves a high abstraction overhead due to boxing).

```
[info] TopPBenchmark.topP  filterAndScan  thrpt    5  28793.391 ± 5723.854  ops/s
[info] TopPBenchmark.topP  filterAndSort  thrpt    5   2531.715 ±   50.438  ops/s
[info] TopPBenchmark.topP        sorting  thrpt    5    132.780 ±    5.378  ops/s
```

karpathy:
Thank you for a nice PR!

Majdoddin:
please also consider #313, constant cut-off. 
The code can safely fallback to other algorithms if constant cut-off doesn't work, but practically that doesn't happen.

PR #284, opened 2023-08-13
multiquery support add

karpathy:
bigchange: add multiquery support in run.c. we can now train and inference multiquery models (where n_kv_heads < n_heads). this also means that we, in principle, support Llama 2 34B and 70B models, which are multiquery


karpathy:
lol i screwed up and pushed this on the wrong branch, my bad. will merge from here anyway

PR #286, opened 2023-08-14
[Feat]: Add support for meta llama hf model conversion

Nick-infinity:
Description:
Llama 2 hf models have weights stored with diff name

### DIFF
--- a/export_meta_llama_hf_bin.py
+++ b/export_meta_llama_hf_bin.py
@@ -1,0 +1,113 @@
+"""
+This script exports the Llama 2 weights in llama2c.bin format.
+"""
+import os
+import sys
+import struct
+from pathlib import Path
+import json
+
+import torch
+
+from model import precompute_freqs_cis
+
+
+def export(p, state_dict, filepath='model.bin'):
+    """export the model weights in fp32 into .bin file to be read from C"""
+    f = open(filepath, 'wb')
+
+    def serialize(key):
+        print(f"writing {key}...")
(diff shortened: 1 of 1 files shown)


Nick-infinity:
Just parking it here for others. Merge is not required

karpathy:
I'll take it because i expect a lot of finetunes will be in HF format. Ty.

mukel:
The HF models apply a permutation to k and v, so the RoPE is different fro the original Llama 2 models. The permutation must be undone to export from HF to llam2a.c's simple .bin files.
It took me quite some time to figure this out, the symptom is it starts degenerating and printing gibberish after ~20 or more tokens.

Nick-infinity:
Yes , I have noticed the gibberish too. That's why I left the comment in
the script for RoPe encodings extraction.

Thanks for validating it

karpathy:
... PR fix welcome

calvintwr:
I don't think this is working. https://github.com/karpathy/llama2.c/issues/314

atamurad:
> The HF models apply a permutation to wq and wk, so the RoPE is different fro the original Llama 2 models. The permutation must be undone to export from HF to llam2a.c's simple .bin files. It took me quite some time to figure this out, the symptom is it starts degenerating and printing gibberish after ~20 or more tokens.
> 
> See https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/convert_llama_weights_to_hf.py#L113-L115

@mukel thanks for tracking this down, fixed it here: https://github.com/karpathy/llama2.c/pull/326

PR #290, opened 2023-08-14
Optimize matices memory arrangement

dror-cyata:
This PR zips the various matrices ([Wq & Wk & Wv] for non-multiquery and [Wk & Wv] for multiquery, as well as [w1 & w3]). This allows multiplication of theses matrices with the relevant vectors in a more efficient manner (both because memory arrangement and because operations on the same vector element are in row).
Bottom line, we get performance improvement from 3 sources:

- Memory arrangement is more efficient
- Computation is more efficient due to vector element caching
- Multithreaded is more efficient, as each task becomes larger, so we have less overhead (more relevant to smaller models)

The following improvements were measured on an M2 machine (clang was used):

With no omp (built with -Ofast -march=native)

| Model | Master | Optimized | Change |
|--------|--------|--------|--------|
| stories15M.bin | 690 | 712 | 3% |
| stories42M.bin | 270 | 283 | 5% |
| stories110M.bin | 100 | 110 | 10% | 

With omp, OMP_NUM_THREADS=4 (built with -Ofast -fopenmp -march=native).

| Model | Master | Optimized | Change |
|--------|--------|--------|--------|
| stories15M.bin | 594 | 712 | 19% |
| stories42M.bin | 298 | 345 | 15% |
| stories110M.bin | 138 | 152 | 10% | 

We also measured the improvements on a Windows machine (i7-8650U, 4 (8 hyperthread) cores). The improvement there was less dramatic: ~5% for stories110M and ~7% for stories15M (both tested with omp)

### DIFF
--- a/run.c
+++ b/run.c
@@ -55,6 +55,9 @@ typedef struct {
     float* freq_cis_imag; // (seq_len, head_size/2)
     // (optional) classifier weights for the logits, on the last layer
     float* wcls;
+
+    float* zipped_wqkv;
+    float* zipped_w13;
 } TransformerWeights;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


jrudolph:
I tried something of a similar kind in llama2.scala, see https://github.com/jrudolph/llama2.scala/blob/main/c/experimental.c. There I just tried to calculate the vector product simultaneously on multiple rows (with the same idea of reusing the input vector for longer). I didn't find any noticable speedup when done AVX2 on my machine, but I guess it's easy to get bottlenecked in different places by changing some details.

Regarding your first point regarding memory arrangement, I wonder if that is really relevant? Assuming the original weights are aligned in memory, cache lines should be smaller than the row size, so it should not matter where you fetch them from. Do you have a theory why that would be faster (e.g. automatic prefetching in the CPU being able to better predict next memory accesses to caches)?

One way you could test this is by avoiding the zipping and just passing multiple input weight arrays to the multiplication functions.

I agree with your assessment of the potential multithreading benefits, the short task durations are not optimal for multithreading.

dror-cyata:
Checking behavior with simpler implementation (no zipping)

PR #293, opened 2023-08-14
Chat UI/UX Server + HTML

musadac:
I have tuned run.c and converted it into web socket  that can be used with html i.e chat.html to interact with the model in website. I will be incorporating more functionality such as:

- Creating Setting Fields that can change parameters such as max_length, temperature etc.
- For Chat I will be incorporating history of chat and how it will be passed to the model.
- 
Screenshot 2023-08-14 at 5 21 51 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+llama2_7b.bin
+stories15M.bin
+chat_run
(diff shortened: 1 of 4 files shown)


karpathy:
This will be nice, especially once we have the Chat models loading and working.

musadac:
> This will be nice, especially once we have the Chat models loading and working.

Can you give me directions on what to work on? Should I work on int4/8 quantization. Then come back to this.

PR #294, opened 2023-08-14
Lighter refactoring. Open up for creatig a "chat" mode.

rdentato:
This proposal keeps the  main() function in run.c structurally identical (i.e. all the variables are declared in the main, no addition of extra types, etc).
If something like this can be implemented,i it will be easier to create a chat (with history, the ability to change parameters from one prompt to another etc).

Just a proposal to help deciding which way to go.


karpathy:
Yes I think we'll need to do a refactor like this a bit in run.c.  This change is also intertwined with the other PR that tries to create an `llm` object, so that the code can be used as a library.

rdentato:
That proposal was mine but I thought it was considered too invasive. 
I do believe that one is cleaner, let me know if there is any good time to resubmit.

PR #297, opened 2023-08-15
Add UTF-8 support to prompts

karpathy:
Almost there. Based on PR from @atamurad , with some changes.


karpathy:
@atamurad thank you for writing tests i appreciate it. I decided not to merge them because of all the .so complications and I couldn't think of a better simple way. I verified a few strings and they all matched and so have you, and hopefully we won't touch this code too much going forward, so just merging like this for now.

PR #300, opened 2023-08-15
Added utf8 validation.

rdentato:
Current handling of UTF-8 does not validate the encoding which, as it is now, may go on until the end of the text.

Added code to check if the encoding is a proper UTF8 encoding
See my full explanation of UTF8 validation at https://dev.to/rdentato/utf-8-strings-in-c-2-3-3kp1

NOTE 1: please advice on how to test with some models whose vocab has non-ASCII character. This need to be tested more before being (in case) accepted.

~~NOTE 2: This patch accepts 0xC080 as a representation of embedded '\0'  (https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8) which might not be needed for llama2.c~~ (Removed, we can add it back if embedded '\0' make any sense in a vocabulary).

NOTE 3: If full validation is overkill, at least there should be a limit at the numbers of byte accepted: no more than four.


karpathy:
Hah ok I didn't realize we have an active UTF-8 resident expert!

- I like the separation of get next utf8 char into a second function
- I don't like that my education comments are gone RIP :'(
- I do like more checks
- I don't have a great testing proposal atm other than eyeballing / trying some strings manually. Maybe we could feed utf8 strings as prompts and verify w.r.t. Python. But that would require UTF-8 vocabs/models. Stories260K currently used in testing is ASCII. Llama 2 is too big for testing...
- Q: does this behavior match what would happen in Python?

rdentato:
I'm pretty sure that Python validates each strings, not sure if it rejects invalid strings or replaces invalid sequences with '�'.
Let's wait for a model with some non ASCII character in it for a thorough test.

xefoci7612:
@rdentato sorry to chime in, I'm no expert of UTF-8, but just to understand, what is wrong with something like this?

```C

int get_next_char(char *text) {

     // Code point ↔ UTF-8 conversion
    // First code point	Last code point	Byte 1	Byte 2	Byte 3	Byte 4
    // U+0000	U+007F	    0xxxxxxx
    // U+0080	U+07FF	    110xxxxx	10xxxxxx
    // U+0800	U+FFFF	    1110xxxx	10xxxxxx	10xxxxxx
    // U+10000	U+10FFFF    11110xxx	10xxxxxx	10xxxxxx	10xxxxxx

    int mask[4] = { 0x80, 0xE0, 0xF0, 0xF8 };

    int len;
    // find how many bytes should be the char
    for (len = 1; len < 5; len++) {
        if ((*text & mask[len-1]) == 0)
            break;
    }

    if (len == 5) {
        // first byte is not among the possible patterns
        return -1;
    }

    // verify all the bytes after the first are correct
    for (int i = 1; i < len; i++) {
        // in a UTF-8 character any byte but the first has format 10xxxxxx
        if (text[i] == '\0' || (text[i] & 0xC0) != 0x80)
            return -1;
    }
    return len;
}

```

rdentato:
Just fixed a bug. 
Let's test with some model with UTF8 encoding before deciding upon merging.

xefoci7612:
@rdentato thanks a lot for the explanations and the link, indeed UTF-8 is a fascinating subject.

In our case we have always the vocab look up downstream, that performs a kind of validation too, indeed the final validation: if is not in the vocab is unknown.

IOW it would make sense for you to split the job in 2 parts? Where get_next_cha() makes a **syntactic parsing** checking that patterns are ok, and the **semantic validation** is left to the vocab look up downstream.

rdentato:
@xefoci7612, I believe is already split up this way. 
The get_next_char() function only checks for "syntactic" validity of the codepoint. 
In the following of the bpe_encoding() function the code point is checked against the vocabulary and, if not found, is split back into individual bytes.
At least, that is my understanding, @karpathy can tell us if I got it right.

rdentato:
@karpathy I have not touched the bpe_encode function but I believe utf8 is not managed correctly,

Imagine you have a model where the name of Japan appears in kanji: "日本"
As the code is now, you will first check that 日 is in the vocabulary and it's very likely it is there since it means "day" (or "sun") and will add this token, instead of looking for the token assigned to Japan: "日本".
So, I believe the strategy should be:

- form a "word" (i.e. a sequence of codepoints separated by some separator)
- if they are not in the vocabulary add tokens for the single codepoint
- if one of the encoding is invalid, add the tokens for the single bytes.

This seems one of those cases where things get messy when moving out of ASCII (which basically is meant to represent English text).

rdentato:
btw, @karpathy, creating some very small model wiith utf-8 characters would be a very good tutorial to show how to train a model with llama2.c.
I would not embark in complex languages like Japanese or Chinese (😨) but doing something in French, Spanish or Italian could be easier (I can help with Italian 😄)

kroggen:
This is from SQLite: (highly tested, public domain)

```c
static const unsigned char sqlite3Utf8Trans1[] = {
  0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
  0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
  0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
  0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
  0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
  0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
  0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
  0x00, 0x01, 0x02, 0x03, 0x00, 0x01, 0x00, 0x00,
};

/* 
** It assumes the UTF8 string is always zero-terminated.
*/
unsigned int sqlite3Utf8Read(
  const unsigned char **pz    /* Pointer to string from which to read char */
){
  unsigned int c = *((*pz)++);
  if( c>=0xc0 ){
    c = sqlite3Utf8Trans1[c-0xc0];
    while( (*(*pz) & 0xc0)==0x80 ){
      c = (c<<6) + (0x3f & *((*pz)++));
    }
    if( c<0x80
        || (c&0xFFFFF800)==0xD800
        || (c&0xFFFFFFFE)==0xFFFE ){  c = 0xFFFD; }
  }
  return c;
}
```

rdentato:
If i read it correctly, this function decodes an UTF encoding and returns 0xFFFD (the lozenge with question mark) if it is invalid.
It seems, however, that for one-byte invalid encodings like 0xBF it returns that byte.

For consistency I would have liked more that it always returned 0xFFFD or always the first byte (as I choosed to do) for invalid encodings.

Also, I can't see were it rejects UTF16 surrogates.

We don't really need to decode the codepoint since we'll look for it in the vocabulary in its encoded form, but this is not relevan. We might test if this is faster and pick the one that we like the most. (Of course I'm biased 😃 )

karpathy:
```c
unsigned int sqlite3Utf8Read(
  const unsigned char **pz    /* Pointer to string from which to read char */
){
  unsigned int c = *((*pz)++);
  if( c>=0xc0 ){
    c = sqlite3Utf8Trans1[c-0xc0];
    while( (*(*pz) & 0xc0)==0x80 ){
      c = (c<<6) + (0x3f & *((*pz)++));
    }
    if( c<0x80
        || (c&0xFFFFF800)==0xD800
        || (c&0xFFFFFFFE)==0xFFFE ){  c = 0xFFFD; }
  }
  return c;
}
```

Why do developers write such ugly code? Who is supposed to read this and understand what is even going on?

@rdentato what do you think are the minimal changes to the current code that would be a lot better?

rdentato:
@karpathy, we can get rid of validation for the sake of simplicity. If the encoding is invalid it will not be found in the vocabulary and will be treated as a sequence of bytes. The worst is that it will introduce noise in form of meaningless tokens.
I need to think about being able to "eat up" meaningful text, I can't think of an example on top of my head. May not be a problem.

You can decide to not validate as long as you're sure that nothing "bad" can happen (e.g. buffer overrun, memory boundary trespassing, etc.), the responsibility will stay with the end user, garbage in, garbage out 😀.

I'll craft a PR with what I consider "minimal" and will send it.

rdentato:
PR#360 has been merged, we can get back to this if needed.

PR #306, opened 2023-08-16
minimal protection against invalid UTF8 encoding.

rdentato:
Accepts invalid encodings

- multiple representation for the same character accepted
- stray non-UTF8 bytes accepted

The minimal protection I've added is to ensure that it does not override the str_buffer memory if the number of stray continuations is greater than max_token_length*2+1


rdentato:
I've also added some space to str_buffer in case max_token_length is one.

karpathy:
I like it ty

rdentato:
@karpathy. I'll keep the validation branch in my repo and close the PR #300.
Should you feel the need to go through validation, I'll resume it.

PR #309, opened 2023-08-16
Slightly tighten init logic

ozabluda:
1. Reuse open file. This removes race condition between 2 open of the same file.
2. Remove unused variables.

### DIFF
--- a/run.c
+++ b/run.c
@@ -613,14 +613,10 @@ int main(int argc, char *argv[]) {
         config.vocab_size = abs(config.vocab_size);
         // figure out the file size
         fseek(file, 0, SEEK_END); // move file pointer to end of file
-        const ssize_t file_size = ftell(file); // get the checkpoint file size, in bytes
-        fclose(file);
         // memory map the Transformer weights into the data pointer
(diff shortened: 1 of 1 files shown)


karpathy:
Sorry I'm not sure I understand the race condition or this fix.

ozabluda:
`FILE *file = fopen(checkpoint, "rb");`

File "checkpoint.bin" is replaced any time between fopen() and open()

`fd = open(checkpoint, O_RDONLY); // open in read only mode`

`fd` now refers to the new (i.e. completely different) file.

karpathy:
I think right now I clean up all the resources explicitly and don't leave anything "hanging", even if it's cleaned up by the OS. I'll try to maintain that for now.

ozabluda:
You currently `exit()` from here without cleaning up `mmap()` or succesfull `calloc()`
https://github.com/karpathy/llama2.c/blob/d73b917d3ba7a1d933c65b1f33ecb58fd8d78a92/run.c#L97C9-L97C28

I can preserve `munmap()`, while removing the unnecessary `open()` with the race condition. Maybe it does serve some purpose, like helping with sanitizers, not sure.

PR #159, opened 2023-07-28
llama2.cu

ankan-ban:
Add simple cuda implementation for llama2 inference
* < 750 lines of code. Idea is to keep it as simple as possible.
* Decided to use FP16 to make llama-7b fit on my GPU (original fp32 weights still loaded and converted on the fly).
* ~60 Tokens/second on RTX 4090 for llama-7b-chat model (sequence length of 269)
  * Performance drops as sequence length increases as the MHA kernel is very inefficient.

Other unrelated changes:
* Cherry-picked a pending pull request to add support for chat (much easier to use and test).
* Reverted the memory mapped IO feature as it doesn't work on Windows, and anyway we just need to upload the weights to GPU memory.

I am actually very impressed by the original llama2.c. With OpenMP enabled on my system (AMD r9 5900X, 3200Mhz DDR4 dual channel memory, Windows 10), I get ~1.6 Tokens per second on the 7b model which is ~85% of peak memory bandwidth. So not only the implementation is simple - it's almost as fast as it can possibly get. For small sequence lengths 60 Tokens/s on RTX 4090 is again close to 85% of peak memory bandwidth utilization, so I believe the only way to make this significantly faster is to use weight-quantization techniques.

PR #310, opened 2023-08-16
float16 and 8-bit CUDA implementations

kroggen:
This is based on the work of @ankan-ban on #159

It has 2 implementations in separate files:

* `llama2.cu` uses float16
* `llama2-q8.cu` uses 8-bit quantization

## Example Usage

For float16:

```
git clone -b cuda https://github.com/kroggen/llama2.c
cd llama2.c
wget https://huggingface.co/karpathy/tinyllamas/resolve/main/stories110M.bin
make cuda
./llama2 stories110M.bin
```

For the 8-bit quantization:

```
git clone https://github.com/kroggen/llama2.c
cd llama2.c
wget https://huggingface.co/karpathy/tinyllamas/resolve/main/stories110M.bin

git checkout quantization-q8
gcc quantize.c -o quantize
./quantize stories110M.bin

git checkout cuda
make cuda-q8
./llama2-q8 data.bin
```


rdentato:
Fwiw, I believe it is key to have in the repo a Cuda implementation (and later an OpenCL one and so one).
This will allow to focus the efforts on supporting GPU while keeping consistency with the official `run.c`.
It also will help in refactoring the code so that the logic will be simple and clean while the implementation details will be abstracted away.

mgrabban:
@kroggen can you check rmsnorm_kernel? I think there are bugs on line [108](https://github.com/karpathy/llama2.c/blob/313b5c93eae498b2b6a5f764a030ed619efb8de3/llama2.cu#L108) (only adds `x[index]`, not `x[index] * x[index]`) and line [113](https://github.com/karpathy/llama2.c/blob/313b5c93eae498b2b6a5f764a030ed619efb8de3/llama2.cu#L113) (sums `ss * ss`, not `ss`).

kroggen:
@mgrabban You are right. Good catch! I will fix it

rdentato:
I know I'm annoying but this is exactly why I believe it's beneficial to have this version in the repo.

karpathy:
This is really cool! Wow. Do you have any stats on the 110M, or even better 7B model?

What are your thoughts on how we maintain all the copy paste code between all these different versions?

- run.c
- runq.c
- run.cu
- runq.cu
- run.cl

etc. etc. :\

rdentato:
Keep them aligned, I would push the differences to specific functions like "load_weights" etc.
If you are not opposed to the idea of creating a llm "object" (like I proposed some PR ago), I'll submit a new one.
I'm traveling these days and I'll be back home tomorrow night. I'll try to cook something up on Monday.

Btw, on my machine the previous version of llama2.cu run at more than 12x wrt the CPU version. I'll test the new version (and the int8 one) when home.

kroggen:
Performance with an RTX 3090 on PCIE 4.0 at 24GB/s:

```
fp16: achieved tok/s: 1300.813008
int8: achieved tok/s: 1262.295082
```

Probably because @ankan-ban created an enhanced matmul kernel on the fp16 version that I had to remove to use a slightly simpler one on the int8 version

I was also wondering in using `bfloat16` instead of `half`, as it is more lossless, but not all cards support it. So maybe as a side work.

kroggen:
> What are your thoughts on how we maintain all the copy paste code between all these different versions?
> 
> * run.c
> * runq.c
> * run.cu
> * runq.cu
> * run.cl

In the beginning we could do that copy and paste by ourselves

But that this is the kind of job that an AI coder model like WizardCoder should do. A mostly repetitive task. It could be setup as a CI job, with some test cases to check if the model did something wrong. We would need some good prompts. That could be fun! I just don't know if these models are good at editing, I suspect they are better at writing new code.

But I agree that it is a good idea to have them in separate files. It is good for understanding and also for performance

Regarding the names to use, we must choose:

* the name of the source-code file
* the name of the compiled file
* the command to build (like `make cuda`) for each

The repo may have int8 and int4 quant versions at some time, so just `runq.c` is not sufficient

We can create a separate issue to discuss and select the best names

ankan-ban:
Great job! I think you can get some more performance optimizing the mat_vec_q8_kernel() a bit - by loading multiple int8 elements at a time (I think loading just 4 int8 elements - i.e load size of uint32_t should be enough to get pretty close to max performance).

kroggen:
Hey @ankan-ban, good to see you back here!

If you wanna do it, you can modify this branch and then send a PR to it (on my fork). When the PR is merged, the commit will appear here

```
git remote add bernardo https://github.com/kroggen/llama2.c
git fetch bernardo
git checkout -b cuda2 bernardo/cuda
```

One thing lacking is to apply the `n_kv_heads` to the MHA. I only did for the other parts

karpathy:
(Btw I really want to get around to the CUDA versions but still a lot of the "basics" of the repo are not where I want them to be. I submitted a bunch of refactors yesterday that I thought clean up the repo quite a lot. I'm still continuing that work a bit, and I'm also still thinking through Chat and Quantization and have to get those into a happy state before I can move on to CUDA)

rdentato:
@karpathy , I see your point, that's why I submitted those minimal PR in the hope they can help you moving faster to your desired state.
However not having this cuda version, which @kroggen is actively working on,  makes more complicated to help him converge toward a stable state. Like in the case of the bug that @mgrabban noted earlier.
Working on two separate repository when you try to make code converge is quite complicated.

Of course it's your call. If you see anything that could help let us know.

For example with the two PR I submitted today (the one on avoiding qsort in encode e the other on having a "generate()" function, I believe that adding a simple "chat mode" would be easier.

kroggen:
@ankan-ban Cool! Is the output content good enough with the int4? Karpathy implemented a grouped version, it is on #312.  @atamurad also implemented an int4 quantization using AWQ. You can check it here:  https://huggingface.co/atamurad/llama2-7b-4bit-awq

calvintwr:
I run into `malloc failed` with this. I was using llama-13b. Any idea?

rdentato:
I got the same issue but I only have 16GB of Ram at the moment. I told myself I would have tried with a bigger machine but never did.
How much RAM do you have?

PR #313, opened 2023-08-17
const cuttoff in sample_topp 

Majdoddin:
Refs #246 #276

This works because, practically, the probabilities that get sampled are more than 0.001. Therefore smaller values can be safely filtered (the cuteoff used in #276 evaluates to ~0.00003).
#276 reduces the sizes of the to be sorted array from 32000 to an average of 172, this PR further reduces it to 12.

There are other possible ways to further optimize the code, but as topp is no more a bottleneck, it doesn't affect the token/s much, thus I don't know if it would be interesting.


karpathy:
I like the current version as it is a bit more principled. I don't think the current setting regresses tok/s noticeable already.

PR #315, opened 2023-08-18
Fix vocab_source in sample.py

davidar:
It was looking in the wrong place for the key, so was just printing gibberish when using a custom tokeniser


karpathy:
oops ty

PR #316, opened 2023-08-18
Training: save meta data to config.json

madroidmaq:
Before the official start of each training, some important parameters configured during training are saved in the config.json file, which is convenient for reviewing some information during model training.

Currently, I just add some fields referring to the config.json information in huggingface, and other field information can also be added.

sample data:

```json
{
    "vocab_size": 32000,
    "torch_dtype": "float16",
    "model_type": "llama",
    "dim": 288,
    "num_layers": 6,
    "num_attention_heads": 6,
    "num_key_value_heads": 6,
    "max_seq_len": 256,
    "parameters": 15191712,
    "transformers_version": ""
}
```


train.py
```diff
@@ -192,6 +193,26 @@
...
 
+def generate_config_json():
+    # Directly reference the variables from the script
+    total_params = sum(p.numel() for p in model.parameters())
+    config_values = {
+        "vocab_size": vocab_size,
+        "torch_dtype": dtype,
+        "model_type": "llama",
+        "dim": dim,
+        "num_layer": n_layers,
```
kroggen, line 204: what about `"num_layers"`?

madroidmaq: thanks, fix  typo

karpathy:
hmm. i could see the appeal of this. why not save just the whole `config` object?

madroidmaq:
> hmm. i could see the appeal of this. why not save just the whole `config` object?

@karpathy You are right, we can save the `config` object directly. I originally thought about aligning with the config.json on `huggingface` model.

I still retain the `parameters` field on the basis of the config object, where you can easily see the number of parameters in the model.

The content of the saved config file is roughly as follows:

```json
{
	"out_dir": "out",
	"eval_interval": 2000,
	"log_interval": 1,
	"eval_iters": 100,
	"eval_only": false,
	"always_save_checkpoint": false,
	"init_from": "scratch",
	"wandb_log": false,
	"wandb_project": "llamac",
	"wandb_run_name": "run2023_08_20_01_39_27",
	"batch_size": 128,
	"max_seq_len": 256,
	"vocab_source": "llama2",
	"vocab_size": 32000,
	"dim": 288,
	"n_layers": 6,
	"n_heads": 6,
	"n_kv_heads": 6,
	"multiple_of": 32,
	"dropout": 0,
	"gradient_accumulation_steps": 4,
	"learning_rate": 0.0005,
	"max_iters": 100000,
	"weight_decay": 0.1,
	"beta1": 0.9,
	"beta2": 0.95,
	"grad_clip": 1,
	"decay_lr": true,
	"warmup_iters": 1000,
	"device": "cuda",
	"dtype": "bfloat16",
	"compile": true,
        "parameters": 15191712
}
```

Issue #311, opened 2023-08-17
Stuck on training dataset: Created a PretokDataset with rng seed 42


madroidmaq:
When I try to train the model, I run into some problems, don't know if anyone has the same problem or how should I solve this problem.

When I execute the training code (below), the log will always be stuck on the output of `Created a PretokDataset with rng seed 42`, and there will be no change for several hours.

Below are some key steps I performed along with my device information.

```python
python train.py
``` 

The corresponding output is roughly as follows:

```
(base) jupyter@instance-20230817-103839:~/llama2/llama2.c$ python train.py
tokens per iteration will be: 131,072
breaks down as: 4 grad accum steps * 1 processes * 128 batch size * 256 max seq len
Initializing a new model from scratch
num decayed parameter tensors: 43, with 15,187,968 parameters
num non-decayed parameter tensors: 13, with 3,744 parameters
using fused AdamW: True
Created a PretokDataset with rng seed 42
``` 

The GPU information of my machine is roughly as follows:

```
Thu Aug 17 05:21:51 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA A100-SXM...  On   | 00000000:00:04.0 Off |                    0 |
| N/A   32C    P0    52W / 400W |   1045MiB / 40960MiB |      0%      Default |
|                               |                      |             Disabled |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|    0   N/A  N/A     40510      C   python                           1042MiB |
+-----------------------------------------------------------------------------+
``` 

The CPU information of my machine is roughly as follows:`12 vCPUs, 85GB RAM`

CatTimson:
I have exactly the same issue. 

"Created a PretokDataset with rng seed 42" and then nothing

madroidmaq:
[...]

Issue #296, opened 2023-08-14
Train wont start for custom dataset


ArtBreguez:
I'm trying to train llama to customer support, i'm using a custom dataset that has this shape:

```
[
    {
        "query": "My order hasn't arrived yet.",
        "response": "We apologize for the inconvenience. Can you please provide your order number so we can investigate?"
    },
    {
        "query": "I received a damaged product.",
        "response": "We apologize for the inconvenience. Can you please provide a photo of the damaged product so we can assist you further?"
    },
    {
        "query": "I need to return an item.",
        "response": "Certainly. Please provide your order number and reason for return, and we will provide you with instructions on how to proceed."
    },
    {
        "query": "I want to change my shipping address.",
        "response": "No problem. Can you please provide your order number and the new shipping address you'd like to use?"
    },
]
```

I already tokenized, but when i run train.py the terminal freezes on:

```
tokens per iteration will be: 131,072
breaks down as: 4 grad accum steps * 1 processes * 128 batch size * 256 max seq len
Initializing a new model from scratch
num decayed parameter tensors: 43, with 15,187,968 parameters
num non-decayed parameter tensors: 13, with 3,744 parameters
using fused AdamW: False
Created a PretokDataset with rng seed 42
```

What could be the reason? When training with TintStories it work fine.

twobob:
Yeah. Im kinda in this space too. Figured it might be a me problem.  still undecided. 

Did try to poke around in exactly where the process thought it was by implementing the entire thing in colab and then stepping into the running process, but still did not figure it out beyond my "training has started" message that I injected fired...

have to revist

madroidmaq:
@ArtBreguez you can try this https://github.com/karpathy/llama2.c/issues/311#issuecomment-1682171746 ，it‘s worked for me。

PR #319, opened 2023-08-19
Give better error message in Tinystories data loader

RahulSChand:
There are few open issues of people getting stuck running `train.py` because `tinystories.py` data loader will be stuck in an infinite loop if there are no .bin files in the data directory. Have added an assert statement so that there is a more informative error message rather than training just being stuck.

Relevant issues
https://github.com/karpathy/llama2.c/issues/296
https://github.com/karpathy/llama2.c/issues/311


madroidmaq (changes requested):
Great tip, I just ran into this problem. I suggest that you can prompt the copy under optimization. When you really encounter this problem, you can provide some tips for checking errors. like:

"No bin files found in {bin_dir}, Please check the pretokenize execution result."

karpathy:
thank you for the PR, good add

Issue #314, opened 2023-08-17
Error in converting huggingface models


calvintwr:
I tried to use `export_meta_llama_hf.bin.py`. The first error I encountered was:

```
FileNotFoundError: [Errno 2] No such file or directory: '../Llama-2-13b-chat-hf/params.json'
```

But that's goes away after I rename the `Llama-2-13b-chat-hf/config.json` to `Llama-2-13b-chat-hf/params.json`.

However, the next error hits, which is:

```py
Traceback (most recent call last):
  File "/root/llama2.c/export_meta_llama_hf_bin.py", line 117, in <module>
    load_and_export(model_path, output_path)
  File "/root/llama2.c/export_meta_llama_hf_bin.py", line 105, in load_and_export
    state_dict = concat_weights(models)
  File "/root/llama2.c/export_meta_llama_hf_bin.py", line 78, in concat_weights
    for name in list(models[0]):
IndexError: list index out of range
```

On closer inspection,`models` originate from https://github.com/karpathy/llama2.c/blob/bd182289c596fa6059eb7b3b7c8ccd04b5c90fc3/export_meta_llama_bin.py#L98

and it calls `.glob` to get `.pth` like this:`model_paths = sorted(list(Path(model_path).glob('consolidated.*.pth')))`

But there are no `.pth` files in the huggingface models.

Nick-infinity:
I forgot to mention that we have to conver the pickle file and dump it as consolidated.pth

Sample code

from transformers import AutoTokenizer, AutoModelForCausalLM
import torch

tokenizer = AutoTokenizer.from_pretrained("meta_llama_2_model")
model = AutoModelForCausalLM.from_pretrained("meta_llama_2_model")
model_path = 'consolidate.pth'

# Save the model's state_dict
torch.save(model.state_dict(), model_path)

karpathy:
try now?

calvintwr:
The conversion works now, but I run into Segmentation fault with bins that I export using `version 1` and `version 2` flags. With `version 0` flag, the model does run but outputs gibberish.

PR #326, opened 2023-08-21
Added huggingface model loader/importer to export.py

atamurad:
Summary:
 * Added `load_hf_model()` function to export.py
 * Correctly set `shared_classifier` flag and write classifier weights if exists, in all export functions
 * Added `hidden_dim` to `ModelArgs`
 * Added `--hf` CLI argument to load HF model

Tests:
 * Legacy export with Llama-2-7b-chat-hf model - OK
 * model.py inference with generate() - OK

Fixes:
 * https://github.com/karpathy/llama2.c/issues/314
 * WQ / WK permutation bug reported in comments: https://github.com/karpathy/llama2.c/pull/286#issuecomment-1679066644

### DIFF
--- a/export.py
+++ b/export.py
@@ -303,11 +303,15 @@ def load_hf_model(model_path):
     model.tok_embeddings.weight = nn.Parameter(hf_dict['model.embed_tokens.weight'])
     model.norm.weight = nn.Parameter(hf_dict['model.norm.weight'])
 
+    # huggingface permutes WQ and WK, this function reverses it
+    def permute_reverse(w, n_heads=config.n_heads, dim1=config.dim, dim2=config.dim):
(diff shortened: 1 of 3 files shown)


karpathy:
@atamurad this looks nice to me. is it ready to merge? the only thing i'd maybe add is we should take out transformers from requirements? it feels like a really heavy dependency for a niche import feature that maybe isn't going to be all that widely used for the core repo? Sometimes I just put a note next to the import giving the pip install command or something, so if someone gets an ImportError they see right away how to fix it.
?

atamurad:
@karpathy yeah wasn't sure about including it in requirements.txt, just removed it and added error message. is phrasing ok? 

otherwise I think it's good to merge.

karpathy:
great! I'll aim to merge tonight.

karpathy:
Cool PR thank you!

PR #327, opened 2023-08-21
Added check of generation parameters.

rdentato:
The checks for valid topp, steps, and temperature were missing in the latest code.
I also moved everything closer to the generation loop as, in the future, this will be within the loop for a chat-like version.

SIDE NOTE:
I've used the idiomatic:
```
if ( a < min || max < a)     
```
to immediately signal that the test is to verify that `a` is OUTSIDE the range `min` .. `max.

Similarly I use:
```
if (  min <= a && a <= max)     
```
to make clear that I want `a` to be INSIDE the range.


run.c
```diff
@@ -745,6 +744,16 @@ int main(int argc, char *argv[]) {
     Sampler sampler;
     build_sampler(&sampler, transformer.config.vocab_size);
 
+
+    // Check for sound parameters and fallback to defaults if needed.
+    // This needs to stay close to the generation code since, in the future,
+    // it will be in the "chat loop" (one may change params from on prompts to the other).
+    if (rng_seed <= 0) { rng_seed = (unsigned int)time(NULL);}
+    if (temperature < 0.0 || 1.0 < temperature) temperature = 1.0;
```
ozabluda, line 752: Why is `temperature` clamped to 1.0, especially without any user-visible diagnostics? It also seems that maybe it's better to print error message for `temperature < 0.0`

ozabluda, line 752: Why is `temperature` clamped to 1.0?

PR #328, opened 2023-08-21
Moved the creation of the sorted vocabulary in "build_tokenizer()"

rdentato:
Currently, the vocabulary is sorted every time the `encode()` function is called.
For the current version, it's not an issue since the program exits every time but if the generation code would be called within a loop to have a "chat" version, this qsort will be useless (and detrimental).

I've moved this to the `build_tokenizer()` function as it seems to me that this is an operation that should logically be performed at the time when we set up the tokenizer.


karpathy:
Slight downside here is that sorting the vocab probably creates latency, and if the user isn't going to prompt the model we are paying the price needlessly. Maybe we can allocate it lazily?

rdentato:
Yes, it can be done in "encode()" to ensure it is done only when needed. I'll change it.

karpathy:
Just pushed a very similar commit ty

PR #329, opened 2023-08-21
Moved export_meta_llama_bin.py to new export.py

atamurad:
Follow up to https://github.com/karpathy/llama2.c/pull/326

Summary:
 * Tested updated `export.py` with Meta LLama2-7B model. 
 * Minor note: moved float32 conversion in `serialize_fp32` from numpy to torch, because numpy does not support BFloat16 dtype (Meta model uses it)
 * Let `argparse` handle one of input model arguments required logic & error message:

```
usage: export.py [-h] [--version VERSION] (--checkpoint CHECKPOINT | --meta-llama META_LLAMA | --hf HF) filepath
export.py: error: one of the arguments --checkpoint --meta-llama --hf is required
```


karpathy:
yay!! 🎉  much nicer.

PR #331, opened 2023-08-21
Add a "generate()" function to further refactoring.

rdentato:
Continuing the refactoring, I suggest to abstract away the "generate()" function.

This way, the `main()` function only sets up the parameters and call the steps involved in generating text:

  - build transformer
  - build tokenizer
  - build sampler
  - generate text


karpathy:
merged a very similar version just now.

PR #335, opened 2023-08-21
Remove unneeded check of free(NULL)

ozabluda:
Passing `NULL` to `free()` is totally allowed

### DIFF
--- a/run.c
+++ b/run.c
@@ -799,7 +799,7 @@ int main(int argc, char *argv[]) {
     }
 
     // memory and file handles cleanup
-    if (prompt_tokens != NULL) { free(prompt_tokens); }
+    free(prompt_tokens);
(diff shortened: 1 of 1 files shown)


rdentato:
Consider that the compiler (at least gcc) emits exactly the same code whether or not you have the check for NULL before calling `free()`
```
	movq	%rax, -8(%rbp)
	cmpq	$0, -8(%rbp)
	je	.L2
	movq	-8(%rbp), %rax
	movq	%rax, %rdi
	call	free@PLT
.L2:
	movl	$0, %eax
	leave
```
it checks for the pointer to be not `NULL` nevertheless.

Yes, it's useless (for code generation) but it saves you if you're on a platform not 100% compliant with the standard.

karpathy:
ok cool TIL

PR #336, opened 2023-08-21
set n0 step count after building transformer

twobob:


rdentato:
Why?

twobob:
> Why?

because the code currently sets steps to 0 in the event of -n not being set  or <= 0 
and
    fprintf(stderr, "  -n <int>    number of steps to run for, default 256. 0 = max_seq_len\n");
says we should do otherwise.  which is more reasonable.

twobob:
> I proposed the same in PR#327. @karpathy merged it but decided not to keep `-n 0` as a request to have the max seq length. (at the moment is my last message on the Discord channel) :)

okay. Then here is another simple chance for him to decide that which he would prefer. 
A reasonable person might expect to not have to know how many steps to allow. :)

rdentato:
Yep, sorry I didn't read your first comment in full and did not get it .

If you have a look at my PR, I proposed these check to be closer to the loop that generates the text so to be ready for the "chat mode".  I guess will come to something similar in due time.

karpathy:
pushed fix, closing.

PR #337, opened 2023-08-21
Changed test on topp.

rdentato:
We already checked that it's between 0.0 and 1.0 when it has been set, there's no need to do it again (every time).

### DIFF
--- a/run.c
+++ b/run.c
@@ -667,7 +667,7 @@ int sample(Sampler* sampler, float* logits, float temperature, float topp) {
         // apply softmax to the logits to get the probabilities for next token
         softmax(logits, sampler->vocab_size);
         // we sample from this distribution to get the next token
-        if (topp <= 0 || topp >= 1) {
+        if (topp == 0) {
             // simply sample from the predicted probability distribution
             next = sample_mult(logits, sampler->vocab_size);
         } else {


ozabluda:
Does it actually make logical sense? if `topp == 0`, logically it's the same as greedy.

rdentato:
Yes, currently the test is:
```
        // we sample from this distribution to get the next token
        if (topp <= 0 || topp >= 1) {
            // simply sample from the predicted probability distribution
            next = sample_mult(logits, sampler->vocab_size);
        } else {
            // top-p (nucleus) sampling, clamping the least likely tokens to zero
            next = sample_topp(logits, sampler->vocab_size, topp, sampler->probindex);
        }
```
but topp is within 0.0 and 1.0, meaning that the `sample_mult()` will only be called when `topp == 0`.
Is to avoid having an additional test.

rdentato:
Wait, the test leaves open the possibility that topp == 1.0.  Does it make sense to have topp at 1.0?
Previously topp= 1.0 meant the same as 0.0 ....

ozabluda:
`topp==1.0` totally makes sense and logically means exactly what follows in the current code:

```
// simply sample from the predicted probability distribution
next = sample_mult(logits, sampler->vocab_size);
```

which is exactly what is documented here ("1.0 = off") :
https://github.com/karpathy/llama2.c/blob/ee95b1bf2943f87bc3e2c845bf07a733aca3806b/run.c#L713

The only reason not to use "topp=1.0" to indicate "off" in the actual code is to prevent using topp in combination with topk, which was always the tradition in LLMs.

karpathy:
@ozabluda is correct, closing

PR #338, opened 2023-08-22
make transformer a global variable

Majdoddin:
This simplifies the code, specially struct `ProbIndex` is removed.

A global `transfomer` is safe, as it is the only instance of its type, and like OOP, makes its data accessible to relevant functions. 

Other changes:
`sample_topp()` allocates its array as a local var (removing `build_sampler()` and `free_sampler()`)
using `stdint` types (still not complete, see #215)

Next Step:
Making tokenizer a global var, for same reasons.

### DIFF
--- a/run.c
+++ b/run.c
@@ -688,15 +688,13 @@ int sample(Sampler* sampler) {
         for (int q=0; q<transformer.config.vocab_size; q++) { transformer.state.logits[q] /= sampler->temperature; }
         // apply softmax to the logits to get the probabilities for next token
         softmax(transformer.state.logits, transformer.config.vocab_size);
-        // flip a (float) coin (this is our source of entropy for sampling)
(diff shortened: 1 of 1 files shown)


karpathy:
Ty but I don't like it. I don't like global varibles too much, I like self-contained units that become lego blocks that can be arbitrarily and safely stacked or composed later.

Majdoddin:
@karpathy ty for reviewing.
The code has got way too many functions.
Making a _unique_ `transformer' global (instead of passing it to every function) does not hurt modularity/reusability.

Formal rigidity bloats the code , a vivid example is having to use `ProbIndex[]` as a filed of `Sampler`, just because `compare()` of `qsort()` does not have access to `transformer`,  which makes  `build_sampler()` and `free_sampler()` necessary.

karpathy:
fair

PR #343, opened 2023-08-23
attempt at chat function

karpathy:
just uploading partial work of the chat function, version 1
not sure i'm super happy with the layout of it just yet
weird hard-coded buffer sizes
seems to not be obviously broken though, e.g. on my machine:

```
ubuntu:~/llamac$ make runomp
ubuntu:~/llamac$ OMP_NUM_THREADS=16 ./run llama2_7b_chat.bin -m chat -y "You only respond with just the first single word that comes to mind" -i "America"

<s>
[INST] <<SYS>>
You only respond with just the first single word that comes to mind
<</SYS>>

America
[/INST]  Flag
</s>
Enter user prompt: Canada

<s>
[INST] Canada [/INST]  Maple
</s>
Enter user prompt: France

<s>
[INST] France [/INST]  Eiffel
</s>
Enter user prompt: Italy

<s>
[INST] Italy [/INST]  Pizza
</s>
Enter user prompt: Iceland

<s>
[INST] Iceland [/INST]  Glacier
</s>
Enter user prompt: Middle Earth

<s>
[INST] Middle Earth [/INST]  Hobbit
</s>
Enter user prompt: 
```

### DIFF
--- a/run.c
+++ b/run.c
@@ -798,6 +798,9 @@ void read_stdin(const char* guide, char* buffer, size_t bufsize) {
     }
 }
 
+// ----------------------------------------------------------------------------
+// chat loop
+
 void chat(Transformer *transformer, Tokenizer *tokenizer, Sampler *sampler,
           char *cli_user_prompt, char *cli_system_prompt, int steps) {
 
@@ -840,7 +843,7 @@ void chat(Transformer *transformer, Tokenizer *tokenizer, Sampler *sampler,
             }
             // render user/system prompts into the Llama 2 Chat schema
             if (pos == 0 && system_prompt[0] != '\0') {
-                char system_template[] = "[INST] <<SYS>>\n%s\n<</SYS>>\n\n%s\n[/INST]";
+                char system_template[] = "[INST] <<SYS>>\n%s\n<</SYS>>\n\n%s [/INST]";
                 sprintf(rendered_prompt, system_template, system_prompt, user_prompt);
             } else {
                 char user_template[] = "[INST] %s [/INST]";
--- a/run.c
+++ b/run.c
@@ -800,16 +800,20 @@ void read_stdin(const char* guide, char* buffer, size_t bufsize) {
 
 // ----------------------------------------------------------------------------
 // chat loop
+// I manually inspected the tokens for a few chat conversations compared to
+// python reference and that seemed ok, but this was not thoroughly tested and
+// is not safely implemented, it's more a proof of concept atm.
 
 void chat(Transformer *transformer, Tokenizer *tokenizer, Sampler *sampler,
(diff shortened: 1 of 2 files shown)


karpathy:
Once I validate that all the special tokens and the schema is correctly followed, I think it would make sense to hide all that stuff away and make it a pretty chat.

karpathy:
Another thought: it *might* be possible to re-use the `generate` function for chat. but it would require quite a few changes inside it, not sure if that's the better path. Leaning to no atm.

rdentato:
I think is important to avoid having to reload the model every time.
With CUDA (and, in general, using a GPU) it will be a slow process. It takes almost 2min on my T4 to load the fp16 llama2-7b-chat model.

karpathy:
> I think is important to avoid having to reload the model every time.
> With CUDA (and, in general, using a GPU) it will be a slow process. It takes almost 2min on my T4 to load the fp16 llama2-7b-chat model.

is this comment related to this PR?

rdentato:
Yes, but neither I remember exactly what I wanted to say. Sorry.

rdentato:
Anyway, my biggest concern on this approach is that it has the template used by llama_7_b_chat embedded into the code.
I don't think it should be merged into the master until there's a way to choose the template.
Unless you consider this version temporary and plan to add the ability to use different templates to a later version.

karpathy:
What other templates are there though? Isn't this _the_ template for Llama 2 chat?

rdentato:
What if anyone trains another chat model using its own template? Not necessarily from one of the Meta lama models.
Isn't llama2.c intent to allow people to train (or fine tune) their model and then launch inference on them?

karpathy:
If they use the exact same format as Llama 2 or finetune a Llama 2 chat model then the code is plug and play
If they wish to innovate on their format then they have to develop their own chat code

rdentato:
Ok. I see your point. I still believe it should be made more general but it's something that could wait for when someone will want to introduce a new template.

run.c
```diff
@@ -785,6 +787,108 @@ void generate(Transformer *transformer, Tokenizer *tokenizer, Sampler *sampler,
...
+            token = prompt_tokens[user_idx++];
+        } else {
+            // otherwise use the next token sampled from previous turn
+            token = next;
+        }
+        // EOS (=2) token ends the Assistant turn
+        if (token == 2) { user_turn = 1; }
+
+        // forward the transformer to get logits for the next token
+        float* logits = forward(transformer, token, pos);
```
nikolaydubina, line 875: just for my knowledge, isn't there any problems with not forwarding model through tokens from users prompt? (e.g. things like KV cache and offsets?)

PR #344, opened 2023-08-23
Train vocab in Python

janimo:
This trains the sentencepiece model without needing to install the binaries (spm_train is not part of the pip package).


karpathy:
oh this is pleasant. we can also delete the train_vocab.sh then

Issue #341, opened 2023-08-22
New export code OOM with 7B model


janimo:
The new export code instantiates a Transformer() model so it needs double the memory (float32 weights vs bfloat16 in the Meta checkpoints).

The old llama only exporter code can easily export the 7B model with 16G RAM (in under 1 minute, not 10 as in the README), the new one gets killed because it's out of memory.

karpathy:
I see, thanks for raising. Thinking...

rogerallen:
Just noting that this runs out of memory on my 32GB system, so it seems to be more than doubling the memory use.

karpathy:
Probably the legacy script export works, I'm guessing?

https://github.com/karpathy/llama2.c/blob/de005474d37d0cde1356739b8c79ebe7b42b5973/export_meta_llama_bin.py

As a temporary patch... sigh

rogerallen:
Indeed, I can confirm that runs in a few minutes (sorry did not time, but ~2 mins) for me. [Python3.8 Ubuntu 20.04 32GB system]

raimondasl:
FWIW, connected to above, new export.py script OOMs with llama-2-70B model on 197G machine. 
llama-2-13B seems to export fine with the same machine.

Running llama-2-13B models exported with --version 2 and --version 1 core dumps:

```
./run llama-2-13b.bin -n 25 -i "Once upon a time"
Segmentation fault (core dumped)

./run llama-2-13b_quantized.bin -n 25 -i "Once upon a time"
Segmentation fault (core dumped)

```
Model exported with --version 0 works fine:

```
./run llama-2-13b_0.bin -n 25 -i "Once upon
 a time"
Once upon a time, the fishing community was known for its strength, courage and determination
```

PR #345, opened 2023-08-23
Added a versione (3) that is like legacy but exports to fp16.

rdentato:
Warning! needs test (I have to little memory. 64MB are required of a 7B parameters model and I couldn't find a smaller one)

### DIFF
--- a/export.py
+++ b/export.py
@@ -69,6 +69,12 @@ def quantize_q80(w, group_size):
     maxerr = err.max().item()
     return int8val, scale, maxerr
 
+def serialize_half(file, tensor):
+    """ writes one fp16 tensor to file that is open in wb mode """
+    d = tensor.detach().cpu().view(-1).to(torch.half).numpy()
+    b = struct.pack(f'{len(d)}e', *d)
+    file.write(b)
+
 # -----------------------------------------------------------------------------
 # legacy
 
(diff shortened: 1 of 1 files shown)


kroggen:
I am against creating a new version. It should export to existing versions, but choosing the output format (fp32, fp16, int8...)

Current code still expect the `freq_cis` to be there in the final file

I also think you mean 64 GB instead of MB. BTW this OOM problem is tracked on #341

rdentato:
It will be the "legacy version" but with f16 weights. 
Because of how `export.py` works, you need to give a version number to it.
We might change the command line parameters of exchange, but this should be the topic of another PR, I guess.

rdentato:
Just tought another way but I don't like it very much: use the extension of the output file to determine the fp32/fp16 size.
For example:
```
python export.py --hf mrm8488/llama-2-coder-7b llama-2-coder-7b.f16
```
could signal `--version 0` (the default) but with weights in fp16 format.  The code would be a little bit uglier but if this could serve to avoid confusion ...

kroggen:
Why your `serialize_half` is not named `serialize_fp16` to be similar with the existing `serialize_fp32`?

Why your function is not together with the other 2 in the beginning of the file?

kroggen:
My attempt is at #347

rdentato:
About the half, is to stay consistent with the torch.half in the code. I liked it better (I ad fp16 initially).

for the position, I don't like to disrupt existing code (unless for something so minimal) and just added at the end of the block. But in this case I agree I should have put it earlier, together with the other two.

PR #347, opened 2023-08-23
export model to fp16

kroggen:


rdentato:
Ok. I see you went for a much  deeper change.

Did you manage to test it?

kroggen:
It is not tested. I am trying to implement the load of model (version 0 and maybe 1)

karpathy:
Question: what is the benefit of fp16?

- As the Llama 2 models were trained in bf16 I find fp16 conversion sketchy. For newly trained models this is less of a concern
- The file sizes are ofc ~2X smaller
- The code is a little bit more bloated

Am I missing some considerations?

export.py
```diff
@@ -31,16 +31,20 @@
...
-    b = struct.pack(f'{len(d)}f', *d)
-    file.write(b)
-
-def serialize_int8(file, tensor):
-    """ writes one int8 tensor to file that is open in wb mode """
-    d = tensor.detach().cpu().view(-1).numpy().astype(np.int8)
-    b = struct.pack(f'{len(d)}b', *d)
+def serialize(file, tensor, type):
+    """ writes one tensor to file that is open in wb mode """
+    if type == 'fp32':
```
karpathy, line 36: feels simplfiable

export.py
```diff
@@ -129,7 +133,7 @@ def legacy_export(model, filepath):
 # -----------------------------------------------------------------------------
 # new version
 
-def version1_export(model, filepath):
+def version1_export(model, filepath, type):
```
karpathy, line 136: we'd have to serialize the type to the header too

rdentato:
The point is that they can be directly loaded into the GPU. 
Not needing conversion on-the-flying (and have a smaller file to load) significantly reduce the load time (which, for my Tesla T4 is around 2min. for the llama2_7b models.

export.py
```diff
@@ -450,6 +454,7 @@ def torchscript_export(model, filepath, zero_params=False, gzip_output=False):
     parser = argparse.ArgumentParser()
     parser.add_argument("filepath", type=str, help="the output filepath")
     parser.add_argument("--version", default=0, type=int, help="the version to export with")
+    parser.add_argument("--type", default='fp32', type=str, help="the data type to export to (fp32, fp16, bfloat16)")
```
karpathy, line 457: i'm not 100% decided if type should be a separate variable that is written into the header, or if it should just be absorbed into version. E.g.:

version 0 original float32
version 1 original float16
version 2 new header int8

etc and just go that way

rdentato: There's another PR that just uses "--version" for this.

kroggen: Don't forget the support to `bf16` (and maybe others to come)

If using a version number for each, it is both not intuitive as also it will have a lot of "versions"

PR #348, opened 2023-08-23
Clip steps maximum value

nehzata:
@karpathy This PR limits the max value of `steps` to not be more than the model's `seq_len` for completeness. It will segfault in the forward pass when `pos` goes beyond `seq_len` otherwise.


karpathy:
ah ok... it's because of the kv cache. in principle could be fixed, there is no real limitation to seq len for the models. ok

nehzata:
> ah ok... it's because of the kv cache. in principle could be fixed, there is no real limitation to seq len for the models. ok

thanks andrej! i'll try to raise a fix PR soon.

PR #350, opened 2023-08-24
Remove script that is no longer used

janimo:


karpathy:
Only problem here atm is that the script has really nice docs about sentencepiece, not sure where to put them. it's a bit too long for making it comments. maybe a new doc.md somewhere... hmmm

karpathy:
done ty

PR #355, opened 2023-08-25
Export vocab size and Code Llama usage docs

janimo:
Code Llama and Code Llama Instruct models use 32016 tokens due to them being additionally finetuned on a fill in the middle task, unlike the Code Llama Python models.

That vocab size needs to be set at export otherwise these models output gibberish.
Some extra work is still needed since code models use RoPE theta of 1000000 instead of 10000 but they do emit code with the vocab_size change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,22 @@ Then chat with it by specifying the chat mode using the `-m` flag, e.g.:
 ./run llama2_7b_chat.bin -m chat
 ```
 
+You can also try Meta's Code Llama models even if support for them is incomplete.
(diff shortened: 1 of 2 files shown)


atamurad:
vocabulary size is missing in `params.json` but I think it should be copied from shape of `tok_embeddings.weight`.

I don't see a reason for the user to provide/override this from CLI option.

janimo:
@atamurad you are right. I thought there may be cases where getting the embedding size is impractical, or there is some reason for having it as an extra field (as seen in HFmodels)  but I guess in our case it makes sense. I have updated the PR. Thanks.

Issue #346, opened 2023-08-23
why not use key and value caches in model.py?


mvuthegoat:
I was just wondering why you didn't use caches to store the key and value tensors in the Transformer like Meta did
Also, Meta uses a different generate function that take advantage of these caches. Their model doesn't have to recalculate K and V with increasing seqlen dimension, like what you did here:
https://github.com/karpathy/llama2.c/blob/7ac65cb2c2b169050747be92011b7bebdd1b4544/model.py#L340-L341

Using caches are clearly more efficient imo, but not sure why they weren't implemented. Am I missing something?

karpathy:
It justs costs lines of code and some complexity that I wanted to avoid. The solution I am more inclined towards is to separate out training and inference, which is done in most serious projects. These two regimes are sufficiently different. So what's missing in this repo is a class usually called Engine, which efficiently serves a trained model. It may be that our sample.py can go on to hold the Engine, and it would maintain the KV cache.

Majdoddin:
theoretically KV cache could be usefull also in training, if `X` increments over a token sequence (`Y` is always the next token).
So from sequence of length `n`, you get `n-1` `(X, Y)` samples.
I think nanoGPT was like that, but I see llama2.c is not like that.

PR #359, opened 2023-08-26
implementin an Engine to serve the trained model by inferencing

Majdoddin:
see #346


Majdoddin:
for each token, the resulting tensors should be concatenated to the cache tensors, which is a costly O(n) operation. 
Is it not better to initialize each cache as [max_new_tokens, ...]?

Majdoddin:
Have just added kv_cache for RMSNorm, to show how it works. 
@karpathy how do you like it?

sample.py
```diff
@@ -34,6 +34,80 @@
...
+        # restore time as batch dimension and concat heads
+        output = output.transpose(1, 2).contiguous().view(bsz, seqlen, -1)
+
+        # final projection into the residual stream
+        output = self.wo(output)
+        output = self.resid_dropout(output)
+        return output
+
+RMSNorm.forward = RMSNorm_forward
+#Attention.forward = Attention_forward
```
Alexey234432, line 107: why is this line commented out? thanks

Issue #332, opened 2023-08-21
Suggestion: Is it possible to reorganize the file structure


madroidmaq:
At present, there are already a large number of source code files in the project (one screen of the browser can't finish displaying), whether it is possible to consider putting the code related to model training under a tired `train` folder, the reasons are as follows:

1. The project is mainly a C reasoning engine, and other logic codes should belong to the second echelon;
2. The code related to the training model is logically separated from the inference engine;
3. It is friendly to beginners/people who are new to this project;

By the way, the same problem exists in the current README.md. There are different structural designs for different groups of people. For the first batch of people, they may just feel the effect of using C language for reasoning. On this basis, those who are interested will further try to train their own models.

The organization I suggest is roughly as follows:
```
├── assets
│   └── llama_cute.jpg
├── train
│   ├── export.py
│   ├── export_meta_llama_bin.py
│   ├── model.py
│   ├── sample.py
│   ├── test_all.py
│   ├── tinystories.py
│   ├── tokenizer.py
│   ├── train.py
│   └── train_vocab.sh
├── LICENSE
├── Makefile
├── README.md
├── build_msvc.bat
├── configurator.py
├── requirements.txt
├── run.c
├── run.ipynb
├── tokenizer.bin
├── tokenizer.model
├── win.c
└── win.h
```

karpathy:
you're not wrong...

karpathy:
Your suggestion isn't quite right either though, e.g. requirements.txt is left outside? and configurator, etc.
And probably should hide away the tokenizer too

rdentato:
@karpaty, some major grouping should be done, I think.
As it stands it's quite confusing.

madroidmaq:
@karpathy I submitted a PR that needs to be merged as soon as possible if this is appropriate, otherwise there may be a lot of git conflicts.

PR #361, opened 2023-08-26
Ref: move train files to train directory

madroidmaq:
See https://github.com/karpathy/llama2.c/issues/332


karpathy:
This PR would break the repo, the move would need careful treatment. I'm not 100% sure it's a good idea yet, let me think

Issue #164, opened 2023-07-28
13B doesn't work for unknown reasons


skeetmtp:
Hello, 

Even if it's low priority, I created this issue to follow investigations.
Feel free to close it, if it's not the way to go.

skeetmtp:
I gave asan a first try
```
rundebug: run.c 
	$(CC) -O1 -g -o run run.c -lm -fsanitize=address -fno-omit-frame-pointer
```

I need to investigate more, but here first result:
```
lempire@spark-lab0:~/alban/llama2.c$ ./run /home/lempire/alban/llama/llama2_13b.bin 1.0 20 "// js function to remove all fields named 'id' from given object\n"
<s>
=================================================================
==76171==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7f14e6469800 at pc 0x556834c5c48a bp 0x7ffedf1e5310 sp 0x7ffedf1e5300
READ of size 4 at 0x7f14e6469800 thread T0
    #0 0x556834c5c489 in matmul /home/lempire/alban/llama2.c/run.c:202
    #1 0x556834c5d525 in transformer /home/lempire/alban/llama2.c/run.c:337
    #2 0x556834c5e9da in main /home/lempire/alban/llama2.c/run.c:550
    #3 0x7f2108d7e082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)
    #4 0x556834c5b56d in _start (/home/lempire/alban/llama2.c/run+0x256d)

0x7f14e6469800 is located 0 bytes to the right of 1677721600-byte region [0x7f1482469800,0x7f14e6469800)
allocated by thread T0 here:
    #0 0x7f21091a8a06 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:153
    #1 0x556834c5b93d in malloc_run_state /home/lempire/alban/llama2.c/run.c:89
    #2 0x556834c5e7b6 in main /home/lempire/alban/llama2.c/run.c:531
    #3 0x7f2108d7e082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)

SUMMARY: AddressSanitizer: heap-buffer-overflow /home/lempire/alban/llama2.c/run.c:202 in matmul
Shadow bytes around the buggy address:
  0x0fe31cc852b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[...]

PR #367, opened 2023-08-27
Do parameter count calculations in 64 bits to not overflow in case of…

janimo:
… very large models

This is based on discussions and code in #111, #154 and #164 so all credits to their authors, and is tested with the 13B and 34B code llama models. It seems to be the minimal amount of conceptual change needed.
The alternative to declaring the new n_layers variable are explicit (unsigned long)p->n_layers casts in 5 of the multiplications.
 
The p->n_layers*p->n_dim could also be precomputed in unsigned long and reused since it appears in all multiplications, but it would lose the pedagogical value of having the order of operations reflect the order of matrices in the cases where dim comes last.

### DIFF
--- a/run.c
+++ b/run.c
@@ -115,26 +115,28 @@ void free_run_state(RunState* s) {
 
 void memory_map_weights(TransformerWeights *w, Config* p, float* ptr, int shared_weights) {
     int head_size = p->dim / p->n_heads;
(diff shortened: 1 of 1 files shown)


kroggen:
Just a reminder that `sizeof(long)` on MSVC is 32-bit

janimo:
@kroggen thank you, updated to unsigned long long.

karpathy:
thank you @janimo

PR #370, opened 2023-08-28
Add llama2.c-web to the list of projects in Readme.md

dmarcos:
I created a simple repo that builds llama2.c with no alterations to WASM and runs it in the Browser.

This is a [demo](https://diegomarcos.com/llama2.c-web/).

- No build system. Just a bash script.
- No Emscripten. 
- [llama2.c repo](https://github.com/karpathy/llama2.c) is a git submodule and code builds without modifications.
- It runs llama2.c in a Web Worker to not block the main thread.
- On my M1 MacBook Air native inference runs at ~100 tokens/s and ~80 tokens/s in the Browser. 20% overhead but haven't spend anytime yet optimizing. Suggestions more than welcome.

I'm also looking for small models that do interesting things and build more Web based demos. Any suggestions super appreciated!

This is a lot of fun. Keep up the great work!


dmarcos:
Thanks for merging. It makes me happy. Open to feedback and willing to keep improving it.

PR #379, opened 2023-09-03
Using Metal allows training on the GPU of Apple devices 

gklab:
Set the parameter device to "mps" to train on Apple devices. Tested on the M2 Ultra chip, the training speed is 6-7 times that of "cpu".

### DIFF
--- a/train.py
+++ b/train.py
@@ -67,7 +67,7 @@ grad_clip = 1.0  # clip gradients at this value, or disable if == 0.0
 decay_lr = True  # whether to decay the learning rate
 warmup_iters = 1000  # how many steps to warm up for
 # system
-device = "mps"  # examples: 'cpu', 'cuda', 'cuda:0', 'cuda:1', 'mps', 'mps:0', 'mps1' etc., or try 'mps' on macbooks
+device = "cuda"  # examples: 'cpu', 'cuda', 'cuda:0', 'cuda:1', 'mps', 'mps:0', 'mps1' etc., or try 'mps' on macbooks
 dtype = "bfloat16"  # float32|bfloat16|float16
 compile = True if device != "mps" else False  # use PyTorch 2.0 to compile the model to be faster.
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -27,10 +27,10 @@
 from model import Transformer, ModelArgs
 from torch.distributed import destroy_process_group, init_process_group
 from torch.nn.parallel import DistributedDataParallel as DDP
-
+import torch._dynamo
 from tinystories import Task
 from export import model_export
-
+torch._dynamo.config.suppress_errors = True
```
karpathy, line 33: not familiar with these, why did you add these two lines? are they safe? importing _modules feels wrong

gklab: > not familiar with these, why did you add these two lines? are they safe? importing _modules feels wrong

Yes, this approach is extreme.
I just adjusted to turn off compilation optimization under MPS. Because the compilation of torch is not complete for MPS, it currently supports backend="aot_eager", but the speed is a little slower than not compiling.

gklab: > not familiar with these, why did you add these two lines? are they safe? importing _modules feels wrong

In addition, the reason why adding these two lines can make mps run normally is because mps will report an error when compiling, and then ignore the compilation behavior. That is to say, mps training is not compiled.

karpathy:
I do want to merge this PR but why is the formatitng so weird? There are many spurious changes here - prints get split up to multiple lines, new lines appear or disappear? If you're able to clean this up to a minimal change that introduces the desired mps behavior without touching anything else I'd be happy to merge.

gklab:
> I do want to merge this PR but why is the formatitng so weird? There are many spurious changes here - prints get split up to multiple lines, new lines appear or disappear? If you're able to clean this up to a minimal change that introduces the desired mps behavior without touching anything else I'd be happy to merge.

Oh, it may be caused by my habitual line wrapping, and part of the reason is the automatic formatting of the IDE.
But this does cause problems for collaborative programming.
I have deleted the redundant code and hope to PR smoothly.

gklab:
> I do want to merge this PR but why is the formatitng so weird? There are many spurious changes here - prints get split up to multiple lines, new lines appear or disappear? If you're able to clean this up to a minimal change that introduces the desired mps behavior without touching anything else I'd be happy to merge.

I've made another adjustment and it shouldn't be simpler.

PR #383, opened 2023-09-05
int8 refactor

atamurad:
This is a refactor PR to be merged to draft PR #364 (branch `feature/int8_try2`):

Summary of changes:
 * `matmul()`, `quantize()` and `dequantize()` all take `QuantizedTensor *` as arguments.
 * Cleaned up `memory_map_weights()`
 * Code does compile and runs as before refactor

### DIFF
--- a/runq.c
+++ b/runq.c
@@ -218,6 +218,16 @@ void build_transformer(Transformer *t, char* checkpoint_path) {
 }
 
 void free_transformer(Transformer* t) {
+    // free QuantizedTensors
+    free(t->weights.token_embedding_table);
+    free(t->weights.wq);
+    free(t->weights.wk);
+    free(t->weights.wv);
+    free(t->weights.wo);
+    free(t->weights.w1);
+    free(t->weights.w2);
+    free(t->weights.w3);
+    free(t->weights.wcls);
     // close the memory mapping
     if (t->data != MAP_FAILED) { munmap(t->data, t->file_size); }
     if (t->fd != -1) { close(t->fd); }
--- a/runq.c
+++ b/runq.c
@@ -38,7 +38,9 @@ typedef struct {
 
 typedef struct {
     // token embedding table
-    QuantizedTensor *token_embedding_table; // (vocab_size, dim)
+    QuantizedTensor *q_tokens; // (vocab_size, dim)
+    float* token_embedding_table; // same, but dequantized
+
     // weights for rmsnorms
     float* rms_att_weight; // (layer, dim) rmsnorm weights
     float* rms_ffn_weight; // (layer, dim)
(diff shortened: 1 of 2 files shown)


xefoci7612:
@atamurad maybe we can run the extra mile and add also the size

```C
typedef struct {
    int8_t* q;    // quantized values
    float* s; // scaling factors
    int n;  // tensor size
} QuantizedTensor;
```


IMHO this simplifies function signatures and the related code even further....

karpathy:
@xefoci7612 one reason to potentially not add n is that it runs cahce alignment? without it the data *might* be cache aligned (?)

runq.c
```diff
@@ -38,23 +38,25 @@ typedef struct {
 
 typedef struct {
     // token embedding table
-    QuantizedTensor token_embedding_table; // (vocab_size, dim)
+    QuantizedTensor *q_tokens; // (vocab_size, dim)
+    float* token_embedding_table; // same, but dequantized
```
karpathy, line 42: is it better to just dequantize embeddings on demand as in my original pr? i don't super love that this way we're adding a lot of memory and latency

karpathy: (startup latency i mean, even though the per token latency would be improved by a tiny amount)

atamurad: Copying my comment from previous PR: 
> In this abstraction - QuantizedTensor is atomic, we do not index into or slice from it. We can matmul with it or dequantize all elements. This abstraction model breaks token embedding (sliced from it) & shared classifier weights (matmul-ed). So I split token embedding into separate rows and exported wcls as one quantizedtensor as well (as if it is not shared). 

So options I could think of are:
1) Partially dequantize QuantizedTensor and break current abstraction of it being opaque/atomic. 
2) Export each row/token embedding as separate QuantizedTensor - this turned out to be bit weird as we had to re-export wcls as 1 QuantizedTensor for final matmul, thus breaking weight sharing (This commit: https://github.com/karpathy/llama2.c/pull/383/commits/f850a97c6ad94b15b8d41fadfa9aa196d615306c)
3) Current approach in this PR - de quantize all at startup. I think this is the best of all options above.

atamurad:
> @xefoci7612 one reason to potentially not add n is that it runs cahce alignment? without it the data _might_ be cache aligned (?)

Why I didn't add it initially is - simply adding n is not enough. We use QuantizedTensor as 1D (activations) and 2D (weights) array interchangeably.

Adding full `shape` array - it is overcomplicating as we don't need more than 2 dimensions. 
Simply adding N (rows) and M (cols) might be a good balance.

I don't see cache alignment as a major issue as we can pad the header.

runq.c
karpathy: I'm not sure that I follow. Why can't we index into the QuantizedTensor into the correct row? This should work just fine as long as dim % GS == 0, which is the case via an assert in the python export code?

atamurad: We can index but goal was to avoid indexing to it from inside `forward()` function (if you want to add support for other quantization methods/techniques, i.e. int4). 
We can implement another `quantized_get_row()` or smth like that separately to decouple quantization from transformer/forward() code.

karpathy: Not sure if I appreciate the int4 issue but maybe I haven't stared at this enough and you're ahead of me. I think I'll merge this for now as it is quite nice.

PR #386, opened 2023-09-07
Using Metal allows training on the GPU of Apple devices

gklab:
[Set the parameter device to "mps" to train on Apple devices. Tested on the M2 Ultra chip, the training speed is 6-7 times that of "cpu".](https://github.com/karpathy/llama2.c/pull/379)
Sir, in the end, only 3 adjustments were made to minimize the impact and allow "mps" training.
I resubmitted a PR.

### DIFF
--- a/train.py
+++ b/train.py
@@ -67,7 +67,7 @@ grad_clip = 1.0  # clip gradients at this value, or disable if == 0.0
 decay_lr = True  # whether to decay the learning rate
 warmup_iters = 1000  # how many steps to warm up for
 # system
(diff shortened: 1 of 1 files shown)


train.py
```diff
@@ -193,6 +194,7 @@
     optimizer.load_state_dict(checkpoint["optimizer"])
 checkpoint = None  # free up memory
 
+compile = compile if device != "mps" else False
```
kroggen, line 197: Does it need to be "pythonic"?

Isn't this clearer?

```python
if device == "mps": compile = False
```

If you guys really prefer "pythonic", there is this option also:

```python
compile = False if device == "mps" else compile
```

PR #397, opened 2023-09-13
Enhance RoPE

### LINKED ISSUE #302: Yet another RoPE rotation block
I was going to send a PR with this change but then saw [your commit](https://github.com/karpathy/llama2.c/commit/a47f9b3969e9f2eb4e41eb177d8c39e33d45153b)

So I am sharing it here instead

It appears to be better for parallelization:

```c
        // RoPE relative positional encoding: complex-valued rotate q and k by freq_cis in each head
        for (int i = 0; i < dim; i+=2) {
            float fcr = freq_cis_real_row[(i % head_size) / 2];
            float fci = freq_cis_imag_row[(i % head_size) / 2];
            // rotate q
            float q0  = s->q[i];
            float q1  = s->q[i+1];
            s->q[i]   = q0 * fcr - q1 * fci;
            s->q[i+1] = q0 * fci + q1 * fcr;
            // rotate k
            if (i < kv_dim) {
                float k0  = s->k[i];
                float k1  = s->k[i+1];
                s->k[i]   = k0 * fcr - k1 * fci;
                s->k[i+1] = k0 * fci + k1 * fcr;
            }
        }
```

It can be made even faster by using `unsigned int` instead of `int` on the loop (the compiler will convert the `/ 2` to `>> 1`, which is faster)

Maybe not so readable as the other code. I am sharing for those who want a little more performance (not benchmarked though).

kroggen:
This makes the RoPE code more clear to understand

It may also be likely faster as it removes the internal loop

Closes #302

### DIFF
--- a/run.c
+++ b/run.c
@@ -268,13 +268,17 @@ float* forward(Transformer* transformer, int token, int pos) {
             float val = pos * freq;
             float fcr = cosf(val);
             float fci = sinf(val);
-            int rotn = i < kv_dim ? 2 : 1; // how many vectors? 2 = q & k, 1 = q only
-            for (int v = 0; v < rotn; v++) {
-                float* vec = v == 0 ? s->q : s->k; // the vector to rotate (query or key)
-                float v0 = vec[i];
-                float v1 = vec[i+1];
-                vec[i]   = v0 * fcr - v1 * fci;
-                vec[i+1] = v0 * fci + v1 * fcr;
+            // rotate q
+            float q0  = s->q[i];
+            float q1  = s->q[i+1];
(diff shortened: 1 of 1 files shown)


karpathy:
This is what it was originally. It creates copy paste, which gives me a bit of anxiety. And it bloats # LOC...

kroggen:
I reviewed the git history using "blame" and I did not find any instance in which it is like this one

The most similar is [this change](https://github.com/karpathy/llama2.c/commit/0447b06e7c30a12ff37d4e34d700065405025439)

You my see it familiar because I wrote a similar code on #302, but it is not a PR

PR #399, opened 2023-09-13
Calculate value directly inside the kv cache

ngc92:
The "value" part of the attention mechanism isn't actually needed until we store it to the kv-cache.
As such, we can just point the output location of `matmul` directly to the position in the value cache, and let the value be calculated directly there.


kroggen:
My attempt is at #400

ngc92:
@kroggen 
yes, you can do the same also with the key. I've updated this.
Note that to me, the main intent of this PR isn't performance (that should be quite neglible), but actually simplicity, because you can get rid of two potentially non-local, dynamically allocated one-letter name variables.

karpathy:
closing due to https://github.com/karpathy/llama2.c/pull/400

PR #400, opened 2023-09-14
calculate key and value inside the kv cache

kroggen:
Less memory allocation and less memory copy

I just don't know if it is clear enough for understanding / learning


kroggen:
This PR is compatible with PR #397. They can be merged without conflicts

PR #410, opened 2023-09-27
Support GQA export, better run.c, Support tinyllama-1.1B

magician-blue:
Add support to tinyllama-1.1B
Add support to convert GQA model (learned from https://github.com/ggerganov/llama.cpp/pull/3364)
Better run.c
- save a little memory(same as #400)
- make rope part a function
- hardcode to check whether there is a `\n`

### DIFF
--- a/run.c
+++ b/run.c
@@ -164,8 +164,8 @@ void read_checkpoint(char* checkpoint, Config* config, TransformerWeights* weigh
 
 // rope_llama(p, s, head_size, pos)
 void rope_llama(Config *p, RunState *s, int head_size, int pos) {
-    int i;
-    #pragma omp parallel for private(i)
+    int i,j;
+    #pragma omp parallel for private(i, j)
     for (i = 0; i < p->n_heads; i++) {
         for (j = 0; j < head_size; j += 2) {
             float freq = 1.0f / powf(10000.0f, (float)j / (float)head_size);


xefoci7612:
Current chat schemas in run.c are based on LLama 2

```C
            // render user/system prompts into the Llama 2 Chat schema
            if (pos == 0 && system_prompt[0] != '\0') {
                char system_template[] = "[INST] <<SYS>>\n%s\n<</SYS>>\n\n%s [/INST]";
                sprintf(rendered_prompt, system_template, system_prompt, user_prompt);
            } else {
                char user_template[] = "[INST] %s [/INST]";
                sprintf(rendered_prompt, user_template, user_prompt);
            }
```

But you may want to use tinyllama's ones instead:

```
<|im_start|>user
Explain huggingface.<|im_end|>
<|im_start|>assistant

```

In general chat templates should be bounded to the loaded pre-trained model, so maybe they should be a configuration parameter in the .bin file

export.py
```diff
@@ -368,11 +368,12 @@ def load_hf_model(model_path):
     config.dim = hf_model.config.hidden_size
     config.n_layers = hf_model.config.num_hidden_layers
     config.n_heads = hf_model.config.num_attention_heads
-    config.n_kv_heads = hf_model.config.num_attention_heads
+    config.n_kv_heads = hf_model.config.num_key_value_heads
     config.vocab_size = hf_model.config.vocab_size
     config.hidden_dim = hf_model.config.intermediate_size
     config.norm_eps = hf_model.config.rms_norm_eps
     config.max_seq_len = hf_model.config.max_position_embeddings
+    config.kv_dim = config.dim * config.n_kv_heads // config.n_heads
```
karpathy, line 376: can this not be derived later in C?

export.py
```diff
@@ -368,11 +368,12 @@ def load_hf_model(model_path):
     config.dim = hf_model.config.hidden_size
     config.n_layers = hf_model.config.num_hidden_layers
     config.n_heads = hf_model.config.num_attention_heads
-    config.n_kv_heads = hf_model.config.num_attention_heads
+    config.n_kv_heads = hf_model.config.num_key_value_heads
```
karpathy, line 371: ?

run.c
```diff
@@ -451,7 +455,12 @@ void safe_printf(char *piece) {
 
 int str_lookup(char *str, TokenIndex *sorted_vocab, int vocab_size) {
     // efficiently find the perfect match for str in vocab, return its index or -1 if not found
-    TokenIndex tok = { .str = str }; // acts as the key to search for
+    char *input = "<0x0A>";
```
karpathy, line 458: why is this delta here done?

export.py
magician-blue: For MHA model, the number of kv heads equals q heads. 
However, for GQA model like llama2-70b, tinyllama1.1B, the number of kv heads and q head are different.

run.c
magician-blue: I'm not sure whether I convert the tokenizer correctly. After I convert the tinyllama-1.1B's tokenizer. The run.c read '<0x0A>' instead of `\n`. I'm trying to figure out how to convert the tokenizer better to remove these line.
Besides, I notice that our run.c can not deal with `\n` in the input (for tinystory 260k,15m,110m model). They will treat it as `\\` and `n`

magician-blue:
> This is cool, I wasn't aware of the TinyLlama 1.1B run. Sounds very nice and useful for this repo to support. Are there any notable architectural changes in it? This PR is a bit of a random combination of **necessary** differences, and a few side optimizations.

There isn't notable architectural changes.

PR #413, opened 2023-10-01
Consider having run.c be split into multiple files

contactandyc:
achieved tok/s: 0.804340 for the 7B dataset using 8 threads on MacBook Pro 2020 32G 13"

I realize that the point of your run.c is that it is Inference Llama 2 in one file of pure C.  For me to understand your code, I found myself breaking it into modules for the primary purpose of making sure that every line was understood. 

See multiple_files/run.c for an example.  All that is left in run.c is the generate, chat, and main functions.  For the main function, parameters are extracted using ak_params.h leaving it reasonably clear how each parameter is used.

I provide an option to compile with PTHREADS=8 to allow a pthread implementation of matmul.  It'd be interesting to see the different C options for doing this both for CPUs and GPUs.  

Each header file has a corresponding c file if there is any complexity to it.  In this case, the c file is still just included at the bottom of the header file to avoid the need for a more complex make system.

I prefixed all exposed functions with ak to help make it clear which functions were written locally and which were system functions.  I realize that the pure C way to do bools is with 0/1 - but the stdbool.h provides true/false and in a system that is basically a giant computation, it is nice to see booleans in the 2-3 places where ints were representing true and false.

Overall, the code is very well written and the zero to hero video series really helps make clear almost everything that is happening in the code.

Thought others might like to see what run.c looks like when broken up into multiple files.

In the multiple_files version, I didn't use mmap as it tends to not be well suited to smaller computers like my MacBook Pro 32G


karpathy:
:) too many files too many problems

PR #429, opened 2023-10-17
KV-Cache for python inferece

joey00072:
28.01 sec >> 4.7 sec
Screenshot 2023-10-17 at 9 38 00 AM


karpathy:
Totally understand, but noone actually wants to sample from the training code. Let's keep it clean and do all the inference stuff in .c code. Ty!

PR #459, opened 2023-12-09
fix allocation of scaling factors

moritztng:
allocate only one scaling factor per group


karpathy:
huh. i'm only doing a quick skim atm. did i mess up the sizing of this oops

moritztng:
you allocate dim elements for the quants and also dim elements for the scaling factors. but there are only dim / gs scaling factors :)

PR #468, opened 2023-12-28
Context extension

hongyihuang:
Extending generation beyond context length window by shifting the kv-cache by -1 and rotating the key (RoPE) by -1. Due to KV-cache is in int8, significant quantization noise exist, is still an issue that needs to be fixed but at least functional for now.


hongyihuang:
Please disregard, I meant to merge this into my own master branch...

PR #473, opened 2024-01-22
Added Haskell version of llama2.c

chris-ch:
I noticed there was no Haskell version, so I implemented it for the sake of completeness


karpathy:
wow very cool ty :)

PR #474, opened 2024-01-25
Improvement Suggestion for malloc_run_state Function

Keyvanhardani:
Currently, memory allocation is being done using calloc, and the program exits immediately upon a failure. This could be improved for better robustness and error handling.


karpathy:
Yeah maybe, but in that case it should be globally done?

PR #512, opened 2024-04-27
Use `int` instead of `float` to calculate `weights_ptr`

qianzhangxa:
The `Config` structure consists of 7 fields of `int` type, so we should use `int` instead of `float` when calculating `weights_ptr`.


jameswdelancey:
Haha yea. I noticed that too but the size is the same.

jameswdelancey:
Been in the code a long time.

PR #513, opened 2024-04-27
Windows _fseeki64 _ftelli64 bug fix to load LLaMA2 7B >4GB weight

jameswdelancey:
Currently, large file support in Windows is borked for master branch. This is the minimum change to get LLaMA 7B (26GB) to run in Windows. For files over 4GB, we need to use a different set of API to replace POSIX fseek and ftell.

$ gcc -v
Target: x86_64-w64-mingw32
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 13.2.0 (Rev6, Built by MSYS2 project)


jameswdelancey:
Second commit reverts the first and the third fixes it more cleanly.

PR #525, opened 2024-06-06
Add llama3.np / llama3.cuda port to README.md

likejazz:


jameswdelancey:
These appear to be llama2 ports vs llama3. They need to at least use tiktoken start tokens to work with the base, not 0 like sentencepiece, and if they work with tinystories, i suspect they are using sentencepiece tokenizer. Chat/instruct require a stack of tokens to start dialogue else you will be unimpressed with the output. There is a predictable output at zero temperature for both base/pretrained and chat/inspect that you can use to validate. You can copy this diff for the meat of what you need: https://github.com/jameswdelancey/llama3.c/commit/7815cd305baf6753241cb4fa6da8d14d1902b2db#diff-8935a7a088435e2ddf7315451f07fae16810932fb3a0a5d706a2eead1618af26R850-R854

PR #528, opened 2024-06-26
Simple text

dmahurin:
These changes add support for training with tinyshakesphere (change from llama2.py), and simple blank line separated text.

### DIFF
--- a/tinyshakespeare.py
+++ b/tinyshakespeare.py
@@ -1,0 +1,140 @@
+"""
+Download, preprocess and serve the TinyShakespeare dataset as a DataLoader.
+
+Follows the same interface as the TinyStories dataset.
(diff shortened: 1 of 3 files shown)


xpww:
Hello! Excuse me, I wrote a tinytext.txt of about dozens of lines. 
When I used 
```
python tinyshakespeare.py 
pretokenize and python train.py --dataset=tinyshakespeare

```
, the following error occurred: 
```
assert num_batches > 0, "this split is way too small? investigate." 
```
I just started to use llm. llama2.c can make it run on my own computer, but I don't have enough basic knowledge to quickly start training a large model of my own. 

Could you please provide me with an example of a related tinytext.txt file? Thank you very much!

PR #532, opened 2024-07-10
bugfix: unexpected behavior when hidden_dim % group_size != 0

EmreAdabag:
This fixes two bugs that cause unexpected behavior when the hidden dim isn't evenly divisible by the quantization group size like in Stories42M which has hidden dim 1376 and group size 64.

1. Matmul uses the wrong scaling factors when performing matmul(_, _, _, hidden_dim, _);
2. When quantizing vectors of length hidden_dim the tail-end hidden_dim % group_size elements aren't quantized.

This fix enables inference to be run with quantized models exported by export.py regardless of hidden_dim % group_size. This has been tested with Stories42M and validated against a python implementation of quantized inference. There will be a negligible performance hit caused by smaller group sizes during the matrix multiplication, otherwise the performance of quantized inference should remain unchanged.

Alternatively/additionally, just ensure that hidden_dim % group_size == 0 in export.py.

### DIFF
--- a/runq.c
+++ b/runq.c
@@ -143,14 +143,15 @@ void dequantize(QuantizedTensor *qx, float* x, int n) {
 }
 
 void quantize(QuantizedTensor *qx, float* x, int n) {
-    int num_groups = n / GS;
+    int num_groups = (n + GS - 1) / GS;     // same as ceil(n / GS) but just int math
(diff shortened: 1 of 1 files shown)


EmreAdabag:
Fixing this in export.py
https://github.com/karpathy/llama2.c/pull/533

PR #543, opened 2024-12-06
Add llama3.2.c port to README.md

Dylan-Harden3:
Clone of llama2.c but updated to work with Llama 3.2 1B/3B base and instruct

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,7 @@ If your candidate PRs have elements of these it doesn't mean they won't get merg
   - [llama2.c-android-wrapper](https://github.com/celikin/llama2.c-android-wrapper): by @[celikin](https://github.com/celikin): added JNI wrapper, PoC
 - C
   - [llama3.c](https://github.com/jameswdelancey/llama3.c): by @[jameswdelancey](https://github.com/jameswdelancey): a LLaMA 3 8B Base and Instruct port of this project
+  - [llama3.2.c](https://github.com/Dylan-Harden3/llama3.2.c): by @[Dylan-Harden3](https://github.com/Dylan-Harden3): a LLaMA 3.2 1B/3B Base and Instruct port of this project
 - C++
   - [llama2.cpp](https://github.com/leloykun/llama2.cpp) by @[leloykun](https://github.com/leloykun): a C++ port of this project
   - [llama2.cpp](https://github.com/coldlarry/llama2.cpp) by @[coldlarry](https://github.com/coldlarry): a C++ port of this project


Uzair-90:
Hey Dylan I have a question, any assistance will be highly appreciated. I want to convert DeepSeek-R1-Llama-8B  into .bin format can I use the same export.py for this?

Dylan-Harden3:
@Uzair-90 Maybe? I only ever tested with meta-llama/Llama-3.2-1B. For `export.py` to work the model needs to be loadable with [transformers](https://huggingface.co/docs/transformers/index) and share all the same parameter names as llama. I don't know what the specifics are for their distillation process so not sure if it will work.
You can try:
```bash
python3 export.py DeepSeek-R1-Distill-Llama-8B.bin --hf deepseek-ai/DeepSeek-R1-Distill-Llama-8B
```
If all you want to do is run a model locally check out [lmstudio](https://lmstudio.ai/), or [ollama](https://ollama.com/), these are more general established projects that let you run basically any of the models whereas this one is hard coded for llama.

Uzair-90:
I already tried this and it works like you can make a .bin file from DeepSeek-Distill-Llama-8B but the provided tokenizer.bin file is not compatible I guess I need to figure out what formatting do I need for my tokenizer.bin

Dylan-Harden3:
@Uzair-90 Yeah so looking at the 2 tokenizers [here](https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Llama-8B/raw/main/tokenizer.json) and [here](https://huggingface.co/meta-llama/Llama-3.2-1B/raw/main/tokenizer.json) they seem to have some small differences but I think you can get around them if you are determined.

It looks like its mainly just the special tokens have different IDs (see the `added_tokens` key in both files), but the mergeable ranks/bpe part seems to be the same (`vocab` key).

I believe you need to edit `tokenizer.py` to make the tokenizer.bin have the special token IDs/ranks for deepseek. If you look at the export method in there all its doing is doing is outputting the tokens with there scores to the tokenizer.bin file.

Uzair-90:
Thank you @Dylan-Harden3 really appreciate it I will look into it.

Uzair-90:
@Dylan-Harden3 can you help me with the problem:

	size mismatch for model.layers.31.self_attn.k_proj.weight: copying a param with shape torch.Size([1024, 4096]) from checkpoint, the shape in current model is torch.Size([4096, 4096]).
	size mismatch for model.layers.31.self_attn.v_proj.weight: copying a param with shape torch.Size([1024, 4096]) from checkpoint, the shape in current model is torch.Size([4096, 4096]).
	You may consider adding `ignore_mismatched_sizes=True` in the model `from_pretrained` method
	
	The same problem arises for all 32 layers what specific changes do I need?
	I ma using transformers 4.30.0

Dylan-Harden3:
@Uzair-90 Is this when you run export.py? Please kindly open a Q&A [discussion](https://github.com/Dylan-Harden3/llama3.2.c/discussions/new/choose) in my fork for further questions, I would like Andrej to approve this PR one day and don't want a long unrelated thread to get in the way.

